PR #2, opened 2021-01-23
setup github actions

jw3:
using the quickstart as a starting point


jw3:
https://github.com/actions-rs/meta/blob/master/recipes/quickstart.md

PR #15, opened 2021-02-12
Daemon control

### LINKED ISSUE #12: Daemon control
Control the fapolicyd daemon

### features
- uses d-bus
- can start / stop
- can enable / disable

jw3:
- dbus
- arbitrary unit
- methods
  - start
  - stop
  - enable
  - disable

closes #12


jw3:
Currently an object works as an unchecked proxy by name to a systemd unit.  Object used for context with anticipation that additional state will be needed.

```
o = Daemon(someunit)
o.start()
o.stop()
```

PR #70, opened 2021-03-31
Improve RPM filtering

jw3:
This brings the filtering of RPM entries more inline with what fapolicyd does by
- drop config
- drop docs
- drop dirs
- apply location and extension specific logic


jw3:
https://github.com/linux-application-whitelisting/fapolicyd/blob/v1.0.2/src/library/rpm-backend.c#L179-L264
https://github.com/linux-application-whitelisting/fapolicyd/blob/v1.0.2/src/library/rpm-backend.c#L272-L374

Issue #64, opened 2021-03-22
Finalize how to handle Untrusted vs Unknown files in the UI trust list


dorschs57:
1. Figure out exactly what it means for a file to be untrusted vs unknown.
    * An untrusted file is one that the user has specifically selected as untrusted in the UI, but has not yet been persisted back to the ancillary trust database?
    * An unknown file is one that either a. has a mismatch in its sha or file size or b. was added from the system trust database?
 1. Develop a good way to visualize with the green/yellow/red and T/U/? indicators the final outcome of #1 above.

jw3:
#73 calls for refactor to T/D/U, defined as

- Trusted; in db and matches
- Discrepancy; in db but does not match
- Untrusted; not in db

jw3:
@dorschs57 are we good on this now given the change to T/D in #104 ?

dorschs57:
I'd say we are good to close this ticket.  If we get different feedback in the future we can open a new ticket. 👍

PR #79, opened 2021-04-07
Refactor backend to TDU

### LINKED ISSUE #78: Include file size in discrepancy check

jw3:
- Trusted: exists in db and matches
- Discrepancy: exists in db but size or hash mismatch
- Unknown: does not exist in db
- Adds checking of file size to discrepancy check
- Changes the unknown representation from an underscore `_` to `U`

closes #78 
ref #64


dorschs57 (approved):
This makes sense to me based on our conversations. Untrusted now really becomes a caveat of the UI only.  A file will only be Untrusted when the user has selected a file, clicked the Untrust button, but has yet to deploy the changes.

PR #80, opened 2021-04-07
Refactor Python dirs

### LINKED ISSUE #75: Merge ui project into py
Simplifies several things and the separation here has not proven to have any advantage.

- single pipenv
- simplifies packaging
- simplifies unit testing
- simplifies understanding
  - ui is part of py, along with util, and any other python impl 

Ill compromise on the fact that the bindings build shouldnt need the UI object (yet).  This is not a library, this is an application, and the application is the UI, and the bindings are there to serve the UI.  So factor all that together and the idea of keeping the bindings separate from the UI in this application repo is over engineering it.

jw3:
Merge the ui dir into the py dir and rename py to python.

closes #75


jw3:
rebased

dorschs57 (approved):
Didn't go through this in detail, but I was able to install my env, run the app, and run the tests without issue.  I'm good to merge this.

PR #82, opened 2021-04-09
Issue 81

### LINKED ISSUE #81: Update UI to work with changes to System backend
Update the UI for the backend changes to System that loads the system and ancillary trust from the fapolicyd db.

* update calls to System
* remove the file choosers 
* load ancillary trust data as soon as the page is opened

dorschs57:
Closes #81 

* updates the UI calls for the latest changes to the backend System
* removes the file choosers 
* loads the data when the pages open
* updates the status markings to us T/D/U

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -74,6 +74,25 @@ jobs:
           name: fapolicy-analyzer-package
           path: ${{ steps.build.outputs.virt_env }}/lib/python3.8/site-packages/fapolicy_analyzer*.egg/
 
+  ui_lint:
+    name: UI Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install dependencies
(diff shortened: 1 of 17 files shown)


tparchambault (approved):
I looked over your changes focusing on the non-test files. assuming that you have a much better handle on the unittest/mocks/etc modules than I. I believe things look good from my perspective. Can the trust_file_list contain duplicates? I'm assuming that this object contains the merged system and ancillary dbases. Can the same file be added multiple times to the ancillary dbase?

dorschs57:
It doesn't look like the rust code is doing anything to filter out dublicates. Since its parsing the data directly from the fapolicyd DB, maybe fapolicyd does something to handle duplicates in the trust file.  From the UI stand point if there is a dublicate in there I'd say we keep it so the user has a chance to something about it (aka untrust/delete one of them).  We'll have to smart about when the add files to the ancillary trust list to make sure to check for dublicates too.

jw3:
> Can the trust_file_list contain duplicates?

- fapolicyd cleans duplicates after loading rpm and file sources
- our changesets flatten to a hash map that is keyed on the path

So duplicates in the backend _shouldnt_ happen.

PR #92, opened 2021-04-18
Init Trust DB in container

### LINKED ISSUE #89: A way to initialize fapolicyd trust db
A significant inconvenience of running in a container now is that the fapolicyd trust db does not exist, and there is no way to create it without starting the daemon.

Running the daemon from the image build is not possible, so we end up having to run the container with exec to a shell so that fapolicyd can be started manually, then killed, to initialize the db.

jw3:
Adds some bin tools to prepare container for fapolicyd, eliminating the need to start the daemon once to generate the db.

- Supports creating an empty db or scanning the RPM database.
- Adds init of db to the docker file
- Adds Makefile for building container image; `make run -C scripts/docker`

closes #89


jw3:
```
#! target/debug/examples/trustdb_init --help
trustdb_init 
Initialize a fapolicyd Trust Database Default behavior is to read from the RPM database Empty
database creation is possible using the `--emtpy` flag

USAGE:
    trustdb_init [FLAGS] [OPTIONS] <dbdir>

ARGS:
    <dbdir>    Directory in which to create the database

FLAGS:
        --empty      create an empty database
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --size <size>    Max size, in bytes, of the database backend [default: 104857600]
```

Issue #88, opened 2021-04-15
Handle adding paths with spaces in them


jw3:
Ensure we can add trust through the ui for paths with spaces in them.

see #74

tparchambault:
Performed some testing to determine if paths with embedded spaces is an issue. The answer is 'yes'.

Test: Attempting to add the file: "File: /home/toma/Development/CTC/data space/man from mars.txt"
In Database:
/home/toma/Development/CTC/data space/man from mars.txt
Size: 36
SHA256: 65d11d006895dead7ec4b253544299709e6cf2ecd6f1e28e454f6e20f5170b1c

On File System:
stat: cannot statx '/home/toma/Development/CTC/data': No such file or directory
stat: cannot statx 'space/man': No such file or directory
stat: cannot statx 'from': No such file or directory
stat: cannot statx 'mars.txt': No such file or directory
SHA256: 65d11d006895dead7ec4b253544299709e6cf2ecd6f1e28e454f6e20f5170b1c

jw3:
The fapolicyd bug that is in progress will cause failures once the path gets into the trust db.  So if we can write the trust db and then fapolicyd fails we are correctly handling them on our end.

jw3:
I wonder if we should add a temporary filter due to the fapolicyd bug to prevent files from spaces from getting in there, until the issue in fapolicyd is resolved.

Would be an interesting feature flag type capability we could provide to work around known fapolicyd issues.

So instead of breaking the fapolicyd trust db we would notify the user that we will not allow them to add the file to the db.  That check would only happen if fapolicyd version matches the known versions to fail.

dorschs57:
That sounds like a good way of handling it.  We'd need some way of pulling back an fapolicyd version from the backend.  Then we could simple pop a warning up if they try to add a file with spaces.

Issue #96, opened 2021-04-19
Travis should build IT image


jw3:
Travis builds will need at times features added in the current branch, which havent been pushed to docker hub yet.

Travis will need to build the image prior to integration tests running.

The question remains, who builds and pushes the latest image that goes to docker hub.

PR #106, opened 2021-04-23
Add integration test for filenames with spaces

jw3:
They dont work, but this test will prove they do when they do.

#88


jw3:
Need to manually push the image for travis, but now is not a good time to do that.

need #96

jw3:
now pushed

PR #111, opened 2021-04-28
Add unit tests for the StateManager

tparchambault:
Added pytest based unit tests to primarily exercise the public interface of the StateManager class. Some tests validate internal object state; Eventually these should be removed because they will complicate refactoring. Currently we are far from a need to refactor.


dorschs57:
Looks good to me and 💯 code coverage 👍.   If its ready go ahead and move it out of draft and merge it.

PR #113, opened 2021-05-04
Async bindings for fetching trust

### LINKED ISSUE #112: Fix UI responsiveness during system trust database loads
The UI can become unresponsive during the loading of the System Trust database.  The UI currently spawns another thread to run this data load but the UI can still become unresponsive depending on the timing of things.  This is caused by the python GIL and the call to the backend not releasing the lock while its working.

dorschs57:
I put together an example of the `system_trust` method that will allow for python multithreading.  Its based on the [parallelism](https://pyo3.rs/v0.13.2/parallelism.html) section of the PYO3 docs.  @jw3 see what you think is the best way to integrate this into the bindings.

closes #112

### DIFF
--- a/python/examples/show_system.py
+++ b/python/examples/show_system.py
@@ -1,24 +1,8 @@
 from fapolicy_analyzer import *
-from concurrent.futures import ThreadPoolExecutor
-
-executor = ThreadPoolExecutor(max_workers=2)
-
-s1 = System()
-print("executing 1")
-future_1 = executor.submit(s1.system_trust_allow_threads)
-
-s2 = System()
-print("executing 2")
-future_2 = executor.submit(s2.system_trust_allow_threads)
-
-result_1 = future_1.result()
-result_2 = future_2.result()
-
-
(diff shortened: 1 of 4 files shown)


python/src/app.rs
```diff
@@ -50,6 +50,10 @@ impl PySystem {
             .collect()
     }
 
+    fn system_trust_allow_threads(&self, py: Python) -> Vec<PyTrust> {
```
jw3, line 53: rename this to `system_trust_async`

jw3 (changes requested):
Move the async example into a separate file named `show_system_async.py` and rename the binding function with `_async` suffix and I think its GTG.

python/examples/show_system.py
```diff
@@ -1,8 +1,24 @@
 from fapolicy_analyzer import *
+from concurrent.futures import ThreadPoolExecutor
```
jw3, line 2: Put the async example in a new file

jw3 (approved):

PR #127, opened 2021-05-18
Add quotes to shell cmd to encapsulate arg1

tparchambault:
The argument to subprocess.getoutput() is a shell command, which uses a whitespace delimiter. In order to pass an argument to the stat cmd that contains whitespace that argument needs to be enclosed within quotes.

### DIFF
--- a/python/fapolicy_analyzer/ui/trust_file_list.py
+++ b/python/fapolicy_analyzer/ui/trust_file_list.py
@@ -1,10 +1,12 @@
 import gi
+import re
 
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk, GdkPixbuf
 from os import path
 from events import Events
 from .ui_widget import UIWidget
+from .confirmation_dialog import ConfirmDialog
 
 
 class TrustFileList(UIWidget, Events):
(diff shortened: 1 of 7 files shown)


python/fapolicy_analyzer/util/fs.py
```diff
@@ -3,7 +3,7 @@
 
 
 def stat(path):
-    return subprocess.getoutput(f"stat {path}")
+    return subprocess.getoutput(f"stat \'{path}\'")
```
dorschs57, line 6: You shouldn't need to escape single quotes inside of doubles in python.

tparchambault: It's not a python thing. The shell is interpreting the whole string.

tparchambault: Maybe the escapes aren't needed.

tparchambault:
Added a block of code to filter out paths that contain spaces and notifying the user with an INFO modal dialog.

python/fapolicy_analyzer/ui/trust_file_list.py
```diff
@@ -106,8 +107,39 @@ def on_addBtn_clicked(self, *args):
...
+                    )
+                    dlgWhitespaceInfo.format_secondary_text(
+                        f"{e}, contains whitespace.\nfapolicyd "
+                        + "currently does not support paths containing spaces."
+                        + "\n\nPath will not be added to the Trusted Files "
+                        + "List. (fapolicyd: VTBD)"
+                    )
+                    dlgWhitespaceInfo.run()
+                    dlgWhitespaceInfo.destroy()
+            files = listTmp
```
dorschs57, line 139: My only comment on this is that maybe we compile a list of files we removed and give one warning message.  If they select 6 files with spaces it would be rather annoying to get the popup 6 times in a row.  You could do something like this with list comprehensions
```python
filtered = [f for f in files if not re.search(r"\s", f)]
diff = [f for f in files if f not in filtered]
if diff:
    ... your dialog code
files = filtered
```

python/fapolicy_analyzer/ui/trust_file_list.py
```diff
@@ -106,8 +107,39 @@ def on_addBtn_clicked(self, *args):
...
+                        flags=0,
+                        message_type=Gtk.MessageType.INFO,
+                        buttons=Gtk.ButtonsType.OK,
+                        text="File path contains embedded whitespace.",
+                    )
+                    dlgWhitespaceInfo.format_secondary_text(
+                        f"{e}, contains whitespace.\nfapolicyd "
+                        + "currently does not support paths containing spaces."
+                        + "\n\nPath will not be added to the Trusted Files "
+                        + "List. (fapolicyd: VTBD)"
```
dorschs57, line 135: Also, python supports the triple quote notation for multi line strings if you want.

python/fapolicy_analyzer/ui/trust_file_list.py
```diff
@@ -106,8 +107,39 @@ def on_addBtn_clicked(self, *args):
         )
         fcd.set_select_multiple(True)
         response = fcd.run()
+        fcd.hide()
```
dorschs57, line 110: Should this be `destroy()` instead of `hide()`?

tparchambault: It gets destroy()ed in its original location. The file selection dialog occasionally would remain on-top of the new whitespace info dlg so I wanted to hide it. We still need the fcd instance farther on down the page to get the file list.

python/fapolicy_analyzer/ui/trust_file_list.py
tparchambault: I was aware of that. I prefer to maintain the indents in the source file. I'll let the CI linting dictate my choice.

python/fapolicy_analyzer/ui/trust_file_list.py
tparchambault: I like that; I will do that.

PR #128, opened 2021-05-20
Issue 62 localization of UI strings

### LINKED ISSUE #62: Handle localization of strings in the UI

dorschs57:
This setups up localization of UI strings via the python `locale.gettext` method.  I also added a set of translations to ES.  I'm not 100% sure the translations are right, but it at least proves that things are working.  

I'm using the Babel tool to generate the template files.  With every change to the UI we'll need to run the tool with `pipenv run extract_messages` to update the template.

In order to run the app with translation you also need to compile the translation file with `pipenv run compile_catalog`.

To test things out for this PR:
1. `cd python`
2. `pipenv run compile_catalog` to compile ES translations
3. `cd fapolicy_analyzer`
4. `LANGUAGE=es python -m ui` to run app with ES locale

closes #62

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -93,7 +93,7 @@ jobs:
         id: install_deps
         run: |
           python -m pip install --upgrade pip
-          pip install flake8
+          pip install flake8 flake8-i18n
       - name: Lint with flake8
         run:  |
           cd python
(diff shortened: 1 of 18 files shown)


jw3:
Is the best practice in Python to use a string literal wrapped in `_()` at every call site, or have a "strings" file?

dorschs57:
> Is the best practice in Python to use a string literal wrapped in `_()` at every call site, or have a "strings" file?

I took @jw3 suggestion and added a "strings" file.  I ended up keeping the f-strings where they were in the widget component files.  I felt it was best to keep those close to the logic that was driving them.  Let me know what you guys think.

jw3 (approved):
:+1:

jw3:
Travis failing with docker hub pull rate limit 

>#   toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

PR #133, opened 2021-06-07
Issue 132 add changeset printout

tparchambault:
The state manager maintains a list of applied but not deployed changes. This PR allows the state manager to request a text based list of path/operation pairs from the lower level rust fapolicyd interface layer, which can then be used by the UI layers.


python/fapolicy_analyzer/ui/state_manager.py
```diff
@@ -73,5 +73,16 @@ def __update_dirty_queue(self):
...
+    def get_path_action_list(self):
+        listReturn = []
+
+        # Iterate through the StateManagers Changeset list
+        for e in self.listChangeset:
+            # Each changeset contains a dict with at least one Path/Action pair
+            for t in e.get_path_action_map().items():
+                listReturn.append(t)
+
+        return listReturn
```
dorschs57, line 85: This whole think can be done with python comprehensions 
```suggestion
        return [t for e in self.listChangeset for t in e.get_path_action_map().items]
```

src/trust.rs
```diff
@@ -156,6 +156,14 @@ impl TrustOp {
             }
         }
     }
+
+    fn to_string(&self) -> (String, String) {
```
jw3, line 160: lets rename this to something like `to_pair` and make it a function in the module instead of a member

tparchambault: How about 'to_string_pair' I'm a big fan of Obvious.

python/fapolicy_analyzer/ui/state_manager.py
tparchambault: Done

src/trust.rs
tparchambault: to_pair( &TrustOp). Done.

src/trust.rs
```diff
@@ -197,6 +205,14 @@ impl Changeset {
     pub fn is_empty(&self) -> bool {
         self.changes.is_empty()
     }
+
+    pub fn get_path_action_map(&self) -> HashMap<String, String> {
+        let mut map_path_action = HashMap::new();
```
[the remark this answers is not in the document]
tparchambault: More or less, as suggested. Done.

src/trust.rs
```diff
@@ -197,6 +205,14 @@ impl Changeset {
     pub fn is_empty(&self) -> bool {
         self.changes.is_empty()
     }
+
+    pub fn get_path_action_map(&self) -> HashMap<String, String> {
```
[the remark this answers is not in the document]
tparchambault: The two functions in src/trust.rs have been moved to the module level.

jw3 (approved):
Looks like it needs a `cargo fmt --all`, otherwise all good here :+1:

dorschs57:
Good here too 👍

dorschs57 (approved):

Issue #115, opened 2021-05-06
Integration test RPM release


jw3:
Full round-trip integration testing on the RPM built in #114 

1. Build RPM
2. Install in ctcoss/fapolicyd container
3. Run existing integration tests on the installation

The third one might take some massaging due to some integration tests exercising things from example directories that are not included in the RPM deployment.

jw3:
The installation check in the RPM workflow sanity checks the RPM.

PR #134, opened 2021-06-10
RPM Build

### LINKED ISSUE #114: Build RPM for release
Package RPM
- [x] include python
- [x] include bindings
- [x] include glade resources

### references
- https://github.com/iqlusioninc/cargo-rpm
- https://msehnout.fedorapeople.org/rust-fedora.pdf

jw3:
## Summary

Build an RPM that includes the Rust bindings and the Python dist.  The standard tools for building Python RPMs or Rust RPMs didn't play well with our layout.  So what we have is a two stage process of building a bdist wheel and then using that wheel to build the RPM.  The source RPM includes the wheel as the source file.

This PR includes GitHub actions workflow to build the RPM.  When pushing a tag the RPM will be published as a release, otherwise RPM is built and discarded.  A future PR will add some automated testing immediately after the build stage.

### Wiki pages related to this PR
- https://github.com/ctc-oss/fapolicy-analyzer/wiki/RPM-Packaging
- https://github.com/ctc-oss/fapolicy-analyzer/wiki/Publishing-Releases

### How the RPM is packaged

1. build a wheel
2. use the wheel as the rpm source
3. install the wheel in the rpm build
4. capture the python site-install files

### Build the builder image

From the root of the project

`docker build -t rpmbuilder -f scripts/rpm/Dockerfile .`

### Build the rpm in the builder container

`docker run --rm -it -v /tmp:/output rpmbuilder`

RPMs will end up in `/tmp` on the host machine.

### Test the rpm in a container
`todo` #115

### Installing the RPM
`dnf install -y fapolicy-analyzer-0.0.4-1.x86_64.rpm`

closes #114


dorschs57:
I was able to build and install the RPM 👍

jw3:
Added a wiki page https://github.com/ctc-oss/fapolicy-analyzer/wiki/RPM-Packaging

tparchambault:
As we talked about over the phone, on a Fedora33 box w/the docker command actually a wrapper script for podman, the final mv of the rpm artifacts failed. I added '--security-opt label=disable' to the docker command line:

> `$ docker run --rm -it -v /tmp:/output --security-opt label=disable rpmbuilder`

The above executed w/o issue.
Ref: https://github.com/containers/podman/issues/3683

jw3:
@tparchambault there are a few draft releases published for testing the CI produced RPM.

tparchambault:
Is there a hard requirement for Python 3.8?

```
sudo yum install ./fapolicy-analyzer-9.9.9-0.rc.x86_64.rpm
Error: 
 Problem: conflicting requests
  - nothing provides python(abi) = 3.8 needed by fapolicy-analyzer-9.9.9-0.rc.x86_64

```
I have the Fedora33 default Python 3.9.5 installed.

jw3:
The spec only calls out `python3`.  Not sure where the `3.8` is being introduced to the RPM at...

The problem is reproducible on a fedora:33 container.

jw3:
```
[root@7694d950f8d6 /]# rpm -qpR /dist/fapolicy-analyzer-9.9.9-0.rc.x86_64.rpm 
dbus-libs
ld-linux-x86-64.so.2()(64bit)
ld-linux-x86-64.so.2(GLIBC_2.3)(64bit)
libc.so.6()(64bit)
libc.so.6(GLIBC_2.14)(64bit)
libc.so.6(GLIBC_2.15)(64bit)
libc.so.6(GLIBC_2.17)(64bit)
libc.so.6(GLIBC_2.18)(64bit)
libc.so.6(GLIBC_2.2.5)(64bit)
libc.so.6(GLIBC_2.3.4)(64bit)
libc.so.6(GLIBC_2.4)(64bit)
libc.so.6(GLIBC_2.9)(64bit)
libdbus-1.so.3()(64bit)
libdbus-1.so.3(LIBDBUS_1_3)(64bit)
libdl.so.2()(64bit)
libdl.so.2(GLIBC_2.2.5)(64bit)
libgcc_s.so.1()(64bit)
libgcc_s.so.1(GCC_3.0)(64bit)
libgcc_s.so.1(GCC_3.3)(64bit)
libgcc_s.so.1(GCC_4.2.0)(64bit)
libm.so.6()(64bit)
libpthread.so.0()(64bit)
libpthread.so.0(GLIBC_2.12)(64bit)
libpthread.so.0(GLIBC_2.2.5)(64bit)
libpthread.so.0(GLIBC_2.3.2)(64bit)
python(abi) = 3.8
python3.8
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(FileDigests) <= 4.6.0-1
rpmlib(PartialHardlinkSets) <= 4.0.4-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
rpmlib(PayloadIsZstd) <= 5.4.18-1
rtld(GNU_HASH)
```

jw3:
Looks like this is a virtual dependency being added by the auto dependency mechanism.  Going to turn that off and lock python to v3.8, we don't test any other versions at this time so it doesn't make sense to allow them. The RPM also sets the paths of the installed library to be 3.8 specific from the build environment (not sure if this is right or not but it is what it is at the moment).

http://ftp.rpm.org/max-rpm/s1-rpm-depend-auto-depend.html
https://unix.stackexchange.com/questions/429664/modify-rpm-package-to-remove-dependencies
https://serverfault.com/questions/411444/rpm-set-required-somepackage-0-5-0-and-somepackage-0-6-0

jw3:
> Going to turn that off and lock python to v3.8, we don't test any other versions at this time so it doesn't make sense to allow them.

The right thing to do here was to build a RPM for each target platform (fc32,33,34,...) and let the python version float to the default version on that platform.  If you try to force to a version it can work for simple cases, but once you start installing additional python dependencies with the system package manager you get into conflicts.

The build has been updated to support this approach.  So we end up with Python 3.9 builds on the later versions, this is noted in #140

PR #136, opened 2021-06-10
Issue 49 summarize changes

tparchambault:
Extended a standard Gtk Dialog to encapsulate a list of the changes to be deployed.


tparchambault:
This may be the issue: https://github.com/rust-lang/rust/issues/81654

python/fapolicy_analyzer/locale/fapolicy_analyzer.pot
```diff
@@ -7,9 +7,9 @@
 #, fuzzy
 msgid ""
 msgstr ""
-"Project-Id-Version: fapolicy-analyzer 0.0.4\n"
+"Project-Id-Version: fapolicy-analyzer snapshot\n"
```
jw3, line 10: Wonder if this is being updated when an RPM build is performed...

tparchambault: If that's the case, wouldn't we see the issue if we installed the newly created rpm on build system? I can try it then rollback the system.

python/fapolicy_analyzer/ui/confirm_changes_dialog.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def main():
+    win = DialogWindow()
+    win.connect("destroy", Gtk.main_quit)
+    win.show_all()
+    Gtk.main()
+
+
+if __name__ == "__main__":
```
dorschs57, line 82: I don't think we want any example or testing code to be committed with new component.

python/fapolicy_analyzer/locale/fapolicy_analyzer.pot
jw3: Will verify this in #148 

Non-blocking for this PR

dorschs57:
I don't think we are using the ConfirmationDialog component any more then, so it can be deleted (py, glade, test files, and possibly string file entries).

dorschs57:
Also we need some test coverage on the new ConfirmationChangesDialog.  You should be able to use the old test_confirmation_dialog.py file as a starting point.

python/fapolicy_analyzer/ui/confirm_changes_dialog.py
```diff
@@ -0,0 +1,83 @@
+import gi
+gi.require_version("Gtk", "3.0")
+from gi.repository import Gtk
+import ui.strings as strings
+
+
+class ConfirmInfoDialog(Gtk.Dialog):
```
dorschs57, line 7: For consistency the class name should correspond with the python file name.

python/fapolicy_analyzer/ui/confirm_changes_dialog.py
tparchambault: Done.

python/fapolicy_analyzer/ui/confirm_changes_dialog.py
tparchambault: Done. Will push shortly.

PR #145, opened 2021-06-16
chore: Rename rust directories

jw3:
I have been thinking of renaming the rust `src` directories, adds clarity (imo) by removing the name ambiguity.

Undecided  but this is what it would look like.


jw3:
image

jw3:
We had agreement that this was a decent thing to do, however this PR has become stale.  I'll restart this effort once there is a decent lull.

jw3:
The best way to have laid out these projects probably was Python and bindings at the top level with a "core" Rust project as a crate project.
```
- fapolicy_analyzer
  - fapolicy_analyzer (python)
  - bindings (rust bindings project
  - core (rust core project)
  - Cargo.toml (just a workspace definition)
  - Pipenv
  - setup.py
  - ...
```

jw3:
Addressed in #216

PR #151, opened 2021-06-23
chore: UI cleanup before starting issue-150

dorschs57:
* Moved `get_content` method into the `UIWidget` base class and renamed it `get_ref`
* Changed `UnappliedChangesDialog` buttons to expand for consistency
* Simplified the interactions with the `AnalyzerSelectionDialog`
* Added a test coverage minimum of 90%
* Added some missing tests

### DIFF
--- a/python/.coveragerc
+++ b/python/.coveragerc
@@ -1,0 +1,10 @@
+[run]
+source =
+    ui
+    util
+
+[report]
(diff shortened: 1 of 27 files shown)


jw3:
lets add a `chore:` tag in front of this in preparation for the changelog automation

Issue #150, opened 2021-06-22
Initial fapolicyd rule management screens


dorschs57:
Create the intial rule management screens based on the prototype document.

PR #152, opened 2021-07-02
chore: refactor of the trust_file_list component for better reuse in Issue-150

dorschs57:
Refactors the trust_file_list component into a new base component called searchable_list.  This new base component can then be reused in the development of the policy rules management screens for #150.

### DIFF
--- a/python/fapolicy_analyzer/tests/test_trust_file_list.py
+++ b/python/fapolicy_analyzer/tests/test_trust_file_list.py
@@ -5,17 +5,17 @@ import gi
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk
 from unittest.mock import MagicMock
-from helpers import refresh_gui
 from ui.trust_file_list import TrustFileList
 
 
+_trust = [
+    MagicMock(status="u", path="/tmp/foo"),
+    MagicMock(status="t", path="/tmp/baz"),
+]
+
+
 def _trust_func(callback):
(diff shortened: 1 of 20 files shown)


python/fapolicy_analyzer/tests/test_trust_file_list.py
```diff
@@ -40,25 +40,27 @@ def test_uses_custom_markup_func():
...
-def test_sorting_path(widget):
-    trustView = widget.trustView
-    assert ["/tmp/foo", "/tmp/baz"] == [x[1] for x in trustView.get_model()]
-    trustView.get_column(1).clicked()
-    assert ["/tmp/baz", "/tmp/foo"] == [x[1] for x in trustView.get_model()]
-
+def test_loads_trust_store(widget):
+    widget.load_store(_trust)
+    view = widget.get_object("treeView")
+    assert ["u", "t"] == [x[0] for x in view.get_model()]
```
tparchambault, line 46: Possibly have your asserts test against the contents of the _trust list since that was what was loaded (In the event that in the future, the strings in the _trust list might get modified, or more are added.)

tparchambault (approved):

python/fapolicy_analyzer/tests/test_trust_file_list.py
dorschs57: Good call.  I updated that test.

PR #153, opened 2021-07-09
Automated changelog generation

jw3:
Adding an automated changelog script and workflow to help maintain a changelog for the project with minimal developer interaction.

Parses the git log and formats commits messages after filtering out commits that match specified patterns.

The pattern for excluding a commit is to prefix the commit title with an ignored tag, `tag: My commit title`

## ignored tags
- `build`
- `chore`
- `fix`
- `fixup`
- `ci`
- `doc`
- `test`

TODO

- [ ] - use doc instead of docs


jw3:
After a while of tagging commits I wonder if the tag should be inverted to identify public facing commits... :thinking: 

If that was the case, then you would have a shorter list of tags to wrangle.

- `feature` - added a feature
- `bugfix` - fixed a bug
- `change` - changed something noteworthy

dorschs57:
A bonus to that approach would be the ability group the changes in the changelog based on their tag.

jw3:
> A bonus to that approach would be the ability group the changes in the changelog based on their tag.

Yeah that is nice.

I am going to rework the python script for this and update the documentation in the wiki.  This PR never merged because it never felt right... this has much more promise I think.

PR #165, opened 2021-07-22
Add Edit Session Json Serialization

tparchambault:
New serialization functionality bound to Open/Save/SaveAs UI elements. Open/loads of a Ancillary Trust Database Admin edit session file does not overwrite current session operations.

### DIFF
--- a/python/fapolicy_analyzer/ui/ancillary_trust_database_admin.py
+++ b/python/fapolicy_analyzer/ui/ancillary_trust_database_admin.py
@@ -1,5 +1,6 @@
 import gi
 import ui.strings as strings
+import logging
 
 gi.require_version("Gtk", "3.0")
 from gi.repository import Gtk, GLib
@@ -29,6 +30,7 @@ class AncillaryTrustDatabaseAdmin(UIWidget):
         )
         self.trustFileList.trust_selection_changed += self.on_trust_selection_changed
         self.trustFileList.files_added += self.on_files_added
+        self.trustFileList.files_deleted += self.on_files_deleted
         self.get_object("leftBox").pack_start(
             self.trustFileList.get_ref(), True, True, 0
         )
@@ -38,7 +40,11 @@ class AncillaryTrustDatabaseAdmin(UIWidget):
             self.trustFileDetails.get_ref(), True, True, 0
         )
 
-        stateManager.changeset_queue_updated += self.on_changeset_updated
+        # To update UI elements, e.g. Deploy button
+        stateManager.ev_changeset_queue_updated += self.on_changeset_updated
+
+        # To reapply user trust add/delete requests from opened session file
+        stateManager.ev_user_session_loaded += self.on_session_load
 
     def __status_markup(self, status):
         s = status.lower()
@@ -115,6 +121,10 @@ SHA256: {fs.sha(trust.path)}"""
             untrustBtn.set_sensitive(False)
             self.trustFileDetails.clear()
 
+    def on_files_deleted(self, files):
+        if files:
+            self.delete_trusted_files(*files)
+
     def on_files_added(self, files):
         if files:
             self.add_trusted_files(*files)
@@ -128,18 +138,23 @@ SHA256: {fs.sha(trust.path)}"""
             self.delete_trusted_files(self.selectedFile)
 
     def on_deployBtn_clicked(self, *args):
+        # Get list of human-readable undeployed path/operation pairs
+        listPathActionTuples = stateManager.get_path_action_list()
+        logging.debug(listPathActionTuples)
         parent = self.get_ref().get_toplevel()
         dlgDeployList = ConfirmInfoDialog(parent)
-        dlgDeployList.load_path_action_list(stateManager.get_path_action_list())
+        dlgDeployList.load_path_action_list(listPathActionTuples)
         confirm_resp = dlgDeployList.run()
         dlgDeployList.hide()
 
         if confirm_resp == Gtk.ResponseType.YES:
             try:
                 self.system.deploy()
-            except BaseException:  # BaseException to catch pyo3_runtime.PanicException
+            except BaseException:
+                # BaseException to catch pyo3_runtime.PanicException
                 stateManager.add_system_notification(
-                    "An error occurred trying to deploy the changes. Please try again.",
+                    "An error occurred trying to deploy the changes. "
+                    "Please try again.",
                     NotificationType.ERROR,
                 )
                 return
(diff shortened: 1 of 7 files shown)


jw3 (approved):
Unrelated failure of RPM build, tracking in #169 

:+1:

python/fapolicy_analyzer/ui/state_manager.py
```diff
@@ -83,17 +88,82 @@ def __update_dirty_queue(self):
...
+            cs = Changeset()
+            if e[1] == "Add":
+                cs.add_trust(e[0])
+            elif e[1] == "Del":
+                cs.del_trust(e[0])
+            else:
+                print("Error: Path/Action Pair: "
+                      "Unknown Action: {} {}".format(e[0], e[1]))
+                continue
+            self.add_changeset_q(cs)
```
dorschs57, line 119: Is this method being used anywhere?  The only calls I see to it are in the tests.

python/fapolicy_analyzer/ui/state_manager.py
```diff
@@ -83,17 +88,82 @@ def __update_dirty_queue(self):
...
+
+    def __path_action_list_to_dict(self):
+        """Convert Path/Action list of tuple pairs to dict for json xfer"""
+        dictPA = dict()
+        for t in self.get_path_action_list():
+            dictPA[t[0]] = t[1]
+        logging.debug("Path/Action Dict: {}".format(dictPA))
+        return dictPA
```
dorschs57, line 127: This could be simplified with dict comprehensions like 
```python
def __path_action_list_to_dict(self):
    return {path: action for (path, action) in self.get_path_action_list()}
```
That being said, you could move this one liner into the main method and eliminate this private method all together.

python/fapolicy_analyzer/ui/state_manager.py
```diff
@@ -83,17 +88,82 @@ def __update_dirty_queue(self):
...
+
+    def __path_action_dict_to_list(self, dictPathAction):
+        """Convert Path/Action dict to list of tuple pairs for json xfer"""
+        listPaTuples = list()
+        for e in dictPathAction.keys():
+            listPaTuples.append((e, dictPathAction[e]))
+        logging.debug("PathAction Tuple List: {}".format(listPaTuples))
+        return listPaTuples
```
dorschs57, line 135: Similar to the above method this can become a one liner with comprehensions:
```python
return [(path, action) for (path, action) in dictPathAction.items()]
```

python/fapolicy_analyzer/ui/state_manager.py
```diff
@@ -83,17 +88,82 @@ def __update_dirty_queue(self):
...
+        with open(strJsonFile, "r") as fp:
+            d = json.load(fp, object_pairs_hook=OrderedDict)
+            logging.debug("Loaded dict = ", d)
+            listPA = self.__path_action_dict_to_list(d)
+            logging.debug("StateManager::open_edit_session():{}".format(listPA))
+
+            # Deleting current edit session history prior to replacing it.
+            if listPA:
+                # ToDo: Delete pending ops in the ATDA's embedded TreeView
+                self.del_changeset_q()
```
dorschs57, line 160: This doesn't delete the changesets from the rust side.  Is that coming in a later commit?

python/fapolicy_analyzer/ui/ancillary_trust_database_admin.py
```diff
@@ -156,3 +171,26 @@ def on_deployBtn_clicked(self, *args):
     def on_changeset_updated(self):
         deployBtn = self.get_object("deployBtn")
         deployBtn.set_sensitive(stateManager.is_dirty_queue())
+
+    def on_session_load(self, listPaPrs):
+        # flake doesn't like long lines so...
+        strFunction = "AncillaryTrustAdmin::on_session_load(): "
+        logging.debug(strFunction + "{} {}".format(len(listPaPrs), listPaPrs))
```
dorschs57, line 178: You could try `logging.debug(f"AncillaryTrustAdmin::on_session_load(): {len(listPaPrs)} {listPaPrs}")` and see if that makes it under the max line size.

dorschs57 (commented):
A couple overall thoughts on the design.
1. I'm not sure I see the point in converting changesets to tuples then back to dicts.  It seems like you have a lot of code dedicated to this, but is it buying us anything?  If I'm reading this right, you are getting a list of dicts from the rust code.  Why not just keep it as a list of dicts?
2. Maybe this is coming in a later commit, but I feel like we need to move some of this code out of the Ancillary Trust DB Admin page.  You talked about developing a manager/controller object.  Now might be the time to implement that.  It doesn't have to be this PR were it happens, but we should have another design discussion on this now that I see what you are doing.

python/fapolicy_analyzer/ui/state_manager.py
tparchambault: No. The ATDA should be the primary client of the rust fapolicyd layer IMO. It also allows it to maintain its UI components so they are consistent with the databases.
Similar, to how the StateManager leverages the ATDA's existing code to add Changesets via an event, I'd like to see the ATDA's api extended so the StateManager can request it to delete the existing undeployed 
ops. Whether the ATDA should contain a queue itself, or just keep things in the existing TreeStore is TBD. I'd like to know your thoughts.

python/fapolicy_analyzer/ui/ancillary_trust_database_admin.py
tparchambault: Ok. Will consider that with the next PR, since it is a branch off of this one more or less.

python/fapolicy_analyzer/ui/state_manager.py
tparchambault: It is a utility for packing the queue as you noted.

Path/Action tuples became the de facto standard structure for working with the TreeViews in the Gtk UI library. Dicts serialize easier. And the queue itself contains Changesets. I've got a ticket open to migrate away from Changesets in the queue. Why propagate the dependency, when the StateManager doesn't interact with the rust interface layer. It worked for prior tickets but it's too opaque now imo.

I did make a quick stab at a single controller (while you were on vaca), but It really was just adding another layer between the only client that would be using those interface functions (almost) so I tried to just change my frame of reference to something like, each tab is it's own application. Sorta, rest/endpoint of approach, because the ATDA maintains the current data in the TreeStore associated with the trust list object afaict.

python/fapolicy_analyzer/ui/state_manager.py
tparchambault: In general, there's something to be said for a function name that tells you what it does w/o reverse engineering code imo. Let's see how things shake out once the StateManager queue elements are migrated to a different type other than Changesets.

python/fapolicy_analyzer/ui/state_manager.py
tparchambault: Some of these conversions will go away after we migrate the elements of the StateManager Q.

PR #177, opened 2021-08-02
Python bindings for users and groups

### LINKED ISSUE #168: Bindings for rules and users
- rules
- users
- groups

### LINKED ISSUE #175: Add users and groups to System
The Python System object will provide a function that fetches users and groups.

The System maintaining these lists makes it possible to role them into the events analysis.

### LINKED ISSUE #176: Load log events from the system
The System object will provide a function to load event logs.

This should load a log file given a path, parse the Events from the log, then roll those Events together with the system state to provide an event analysis.

jw3:
- Adds user and group bindings
- Adds loading functions
- Adds users and groups to system state

### todo
- refactor the enum passing through pyo3 #181 

closes #168
closes #175
closes #176


jw3:
```plain
$ python3 examples/show_system.py
WARN: fapolicyd trust db was not found
found 0 system trust entries
found 38 system users
found 88 system groups
```

jw3:
how many users are in the 'adm' group

`print(next(len(g.members) for g in s1.groups() if g.name == 'adm'))`

jw3:
To support the user and groups we will need to recommend the syslog format be changed, config looks like

`syslog_format = rule,dec,perm,uid,gid,pid,exe,:,path,ftype,trust`

A syslog entry would then look like

`Aug 04 12:37:02 24f915ea779f fapolicyd[36]: rule=1 dec=deny_syslog perm=execute uid=10001 gid=12345 pid=85 exe=/usr/bin/bash : path=/deny/ls ftype=application/x-executable trust=0`

dorschs57:
Code looks good and the example script ran for me. 👍

PR #180, opened 2021-08-04
chore: Fixed loading of locales from RPM install

dorschs57:
I based this fix off of this stackoverflow https://stackoverflow.com/a/55498913/4351673.


jw3:
Fixed some dependency wiring in here too

PR #195, opened 2021-08-17
Add System view rollback to the Ancillary Trust Database Admin

tparchambault:
Functional but still needs unit tests and minor refactoring.


tparchambault:
Good to go. Moved functionality into class methods. Easier to unit-test, and easier to move into the StateManager if we collectively feel it's robust enough.
Added a new attribute that represents a System snapshot at either start-up or the last successful deployment of changesets to fapolicyd.

dorschs57 (commented):
This looks pretty clean for how the app is handling the system object now.  Like we've talked about, there will be some changes once we workout the state management of the system, but this is a great starting point.

PR #196, opened 2021-08-20
Removed selector dialog and default to DB Admin page

dorschs57:
* remove selection dialog and default to db admin
* implemented the Action menu
* reset of sort order of trust view on refresh from apply changeset

closes 164

### DIFF
--- a/python/fapolicy_analyzer/ui/ancillary_trust_file_list.py
+++ b/python/fapolicy_analyzer/ui/ancillary_trust_file_list.py
@@ -15,6 +15,7 @@ class AncillaryTrustFileList(TrustFileList):
     def __init__(self, trust_func):
         addBtn = AddFileButton()
         addBtn.files_added += self.on_addBtn_files_added
+        self.changesColumn = None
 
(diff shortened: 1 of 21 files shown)


python/fapolicy_analyzer/ui/ancillary_trust_file_list.py
```diff
@@ -46,17 +47,20 @@ def reducer(map, path):
...
         )
-        statusColumn.set_sort_column_id(4)
-        return [statusColumn, *super()._columns()]
+        self.changesColumn.set_sort_column_id(4)
+        return [self.changesColumn, *super()._columns()]
 
     def load_trust(self, trust):
         changesetMap = self._changesets_to_map()
 
+        # Hide changes column if there are no changes
```
tparchambault, line 61: I think from the occasional user's perspective or when support is being provided virtually, it would make sense to always have it visible for the sake of consistency and expectations. Consider the case of a user searching for a particular display, "the one with the changes column...," then again I just might be that confused user.

tparchambault (approved):

PR #197, opened 2021-08-20
Backup fapolicyd data and config prior to deployment

tparchambault:


tparchambault:
PR adds the backup functionality to the tool. In the Ancillary Trust Database Admin object, after the Deploy button is invoked but prior to system deploy, a global utility function in the fatd module is invoked and blindly captures the contents of /var/lib/fapolicyd and /etc/fapolicyd/ in tarball format. These tarballs are generated regardless of the success or failure of the deploy operation.

The tarballs are stored in the XDG_DATA_HOME directory, with the default location being $HOME/.local/share/fapolicy-analyzer/.

python/fapolicy_analyzer/tests/test_xdg_utils.py
```diff
@@ -0,0 +1,106 @@
+import os
+from util.xdg_utils import (
+    xdg_data_dir_prefix,
+    xdg_state_dir_prefix,
+    xdg_config_dir_prefix,
+)
+
+
+def test_xdg_state_dir_prefix_wo_env(monkeypatch):
```
dorschs57, line 9: The `monkeypatch` argument is not being used in this test.

python/fapolicy_analyzer/tests/test_xdg_utils.py
```diff
@@ -0,0 +1,106 @@
...
+
+def test_xdg_state_dir_prefix_w_exception(monkeypatch):
+    strBasename = "Arbitrary.txt"
+    strEnvVar = "/usr/lib"  # Unwritable system dir
+    strExpected = "/tmp/" + strBasename
+
+    # Mock the environment to use a directory that can not be created
+    monkeypatch.setenv("XDG_STATE_HOME", strEnvVar)
+    assert xdg_state_dir_prefix(strBasename) == strExpected
+    monkeypatch.delenv("XDG_STATE_HOME", raising=False)
```
dorschs57, line 44: Can you create a fixture called `env` to DRY some of this repeated code up?
```python
@pytest.fixture
def env(monkeypatch):
    monkeypatch.setenv("XDG_STATE_HOME", strEnvVar)
    yield
    monkeypatch.delenv("XDG_STATE_HOME", raising=False)
...
 def test_xdg_state_dir_prefix_w_exception(env):
    strBasename = "Arbitrary.txt"
    strEnvVar = "/usr/lib"  # Unwritable system dir
    strExpected = "/tmp/" + strBasename
    assert xdg_state_dir_prefix(strBasename) == strExpected
```

python/fapolicy_analyzer/util/fapd_dbase.py
```diff
@@ -0,0 +1,77 @@
...
+    print("Executing: ", listTarCmd)
+    try:
+        print("Executing: ", listTarCmd)
+        p = subprocess.Popen(listTarCmd)
+        retCode = p.wait()
+
+        # Delete oldest back-up file
+        if retCode == 0:
+            fapd_dbase_cleanup_snapshots(strArchiveBasename)
+            return True
```
dorschs57, line 70: `tarfile` is a standard python library if you want to use that instead of running this as a subprocess.

python/fapolicy_analyzer/util/fapd_dbase.py
```diff
@@ -0,0 +1,77 @@
...
+    if os.path.isfile(strListFile):
+        listTarSourceArgs = ["-T", strListFile]
+    else:
+        listTarSourceArgs = ["/var/lib/fapolicyd/", "/etc/fapolicyd/"]
+
+    # Build the tar command w/source options to either use the list or defaults
+    listTarCmd = ["/usr/bin/tar", "-czf", strArchiveFile] + listTarSourceArgs
+    print("Executing: ", listTarCmd)
+    try:
+        print("Executing: ", listTarCmd)
```
dorschs57, line 63: Duplicate print statements here.  I'd say we could move the prints to `logging.debug` statements too

python/fapolicy_analyzer/util/xdg_utils.py
```diff
@@ -0,0 +1,67 @@
...
+
+def xdg_config_dir_prefix(strBaseName):
+    """Prefixes the file basename, strBaseName, with the XDG_CONFIG_HOME
+    directory, and verifies that it is readable by the effective user.
+    """
+    # Use the XDG_CONFIG_HOME env var, or $(HOME)/.config/
+    _home = os.path.expanduser("~")
+    xdg_config_home = os.environ.get("XDG_CONFIG_HOME", os.path.join(_home, ".config"))
+    app_tmp_dir = os.path.join(xdg_config_home, "fapolicy-analyzer/")
+    return app_tmp_dir + strBaseName
```
dorschs57, line 67:
```suggestion
    return os.path.join(xdg_config_home, "fapolicy-analyzer/", strBaseName)
```

python/fapolicy_analyzer/util/xdg_utils.py
```diff
@@ -0,0 +1,67 @@
+import os
+
+
+def xdg_state_dir_prefix(strBaseName):
```
dorschs57, line 4: This and `xdg_data_dir_prefix` methods seem to repeat a lot of code.  Maybe create a private method to handle it that takes either `state` or `share` as a 2nd argument and call it from these 2 methods.

python/fapolicy_analyzer/tests/test_xdg_utils.py
dorschs57: I never did this before, but I think it would work.

python/fapolicy_analyzer/util/fapd_dbase.py
dorschs57: The comment about changing `print` to `logging.debug` could go for all of them unless they serve a purpose like dumping an exception.

python/fapolicy_analyzer/tests/test_xdg_utils.py
tparchambault: Agreed. Removed.

python/fapolicy_analyzer/tests/test_xdg_utils.py
tparchambault: Other than the env var is set to two different directories to check a successful dir creation, and conversely to check dir creation failure to generate an exception. I don't know how to parameterize a fixture. I'll do some research...

python/fapolicy_analyzer/util/fapd_dbase.py
tparchambault: I am aware of that, and have used it before. I'm not a fan of abstracting standard utilities to have a more complicated interface that needs to be researched unless it is a clear win. E.g. if we were on Windows, or if this tool was cross platform then I would absolutely go that route.

python/fapolicy_analyzer/util/fapd_dbase.py
tparchambault: Cruft from debugging. Will remove... os.path.join() has some nice idiosyncrasies. If any component argument starts with a slash, all component arguments to the left of that argument are discarded.

python/fapolicy_analyzer/tests/test_xdg_utils.py
dorschs57: You can look at how I did the `confirm_dialog` fixture in `test_ancillary_trust_database_admin.py`.

python/fapolicy_analyzer/util/xdg_utils.py
tparchambault: Done

python/fapolicy_analyzer/util/xdg_utils.py
tparchambault: Will create a utility helper function...

python/fapolicy_analyzer/tests/test_xdg_utils.py
tparchambault: There doesn't appear to be a consensus or at least a simple one out in the community. Will continue to look because in the long run it will be useful for our overall unit=test suite.

python/fapolicy_analyzer/util/xdg_utils.py
tparchambault: Done

dorschs57:
Looks good. Merging.

PR #201, opened 2021-08-25
Added switching between ACL and subject first event analysis

dorschs57:


python/fapolicy_analyzer/ui/main_window.py
```diff
@@ -42,6 +42,7 @@ def __init__(self):
 
         # Disable 'File' menu items until backend support is available
         self.get_object("restoreMenu").set_sensitive(False)
+        self.__toggle_trustDbMenu()
```
tparchambault, line 45: The UX loading of event logs might not be obvious if you are already in the Event Analysis Tool, i.e. Action > Analyze From Audit Log. The File > Open convention might be more obvious, or a separate Tool menu.

tparchambault (approved):

PR #232, opened 2021-09-14
Improved trust handling

jw3:
Improving the handling of trust to support rolling trust status into the policy analysis view.

for #235

This does several things
- Moves the trust source out of the api, way down into pub crate for trust
- Adds a container backed by a hash map to manage trust and metadata
- Adds a "U" state, for missing files
- Moves the last modified time under a stat call on the trust object to group together a bunch of optional stat like metadata
- Run unit tests on full workspace in ci
- Refactor and simplify the trust reading machinery
- Adds Rayon to trust loading

### todo
- [x] db unit tests
- [x] update python examples
- [x] update rust doc


crates/trust/src/db.rs
```diff
@@ -0,0 +1,82 @@
...
+        matches!(&self.source, Some(TrustSource::System))
+    }
+
+    pub fn is_ancillary(&self) -> bool {
+        matches!(&self.source, Some(TrustSource::Ancillary))
+    }
+}
+
+#[cfg(test)]
+mod tests {}
```
jw3, line 82: - [ ] db tests

jw3:
Unfortunately I figured out how to do a bulk organize import at 0030, and at that time it seemed like a good time to roll it into this.  So there are a few unrelated files and change sets included.

tparchambault (approved):

jw3:
Modified commit message, new commit is f4a4061d

PR #236, opened 2021-09-16
Add mtime field to trust lists

tparchambault:


tparchambault:
This PR adds the file's last modification time to both the System and Ancillary trust file lists.

jw3:
I wonder if mtime would be better as the right-most column.

tparchambault:
> I wonder if mtime would be better as the right-most column.

Ha!

jw3:
It is still of less significance than the path in all situations.

dorschs57 (approved):

PR #240, opened 2021-09-17
Implement UI state management with Redux

dorschs57:


tparchambault:
Pull the redux branch and gave it a spin. Ran well and correctly for superficial testing. The only concrete issue that I saw was that an attempted deploy from the ATDA tab, and with fapolicyd not running, did not present the "Failed deploy" pop-up. The ATDA presented the infinite spinning balloons gfx.

Otherwise looks good for Open/Save/Restore edit session functionality.

tparchambault:
Quick test of the current redux branch: 
- With one 'add' changeset in the ATDA tab, transitioned to the Analysis tool (loaded the standard event0 log), and then returned to the ATDA. The previously loaded changeset was no longer visible in the ATDA tab. So the question is, do we expect to maintain Trust edit session state.

dorschs57:
> Quick test of the current redux branch:
> 
> * With one 'add' changeset in the ATDA tab, transitioned to the Analysis tool (loaded the standard event0 log), and then returned to the ATDA. The previously loaded changeset was no longer visible in the ATDA tab. So the question is, do we expect to maintain Trust edit session state.

It should maintain that state between screens and reload the changesets into the view when you navigate back to the ATDA.  I'm seeing no issues with that running it here locally.

tparchambault (approved):
I haven't verified the autosaved file generation/detection/restore, and the fapolicyd state saving on deploy, If you've tested them, feel free to merge. Otherwise I'll test them on Monday AM.

tparchambault:
Autosave generation/detection/restore and fapd archives on deploy are good to go.

PR #244, opened 2021-09-20
Cache resolved trust

jw3:
Moves resolution of trust status to system boot time, caching it for reuse.

This eliminates the need for async fetches in the bindings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -256,7 +256,6 @@ dependencies = [
  "fapolicy-daemon",
  "fapolicy-trust",
  "pyo3",
- "rayon",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


jw3:
@dorschs57 The calling of the loading spinner was not updated, do _you_ want to update in this PR?

jw3:
@dorschs57 also would be useful if you could test this against the points you were seeing the repeated delays and confirm that this eliminates those.

dorschs57 (approved):
Some quick testing of these changes on the redux branch proves to me that the caching is working.
1. first call to `system_trust` after initial startup time took 0.17 secs
2. navigate to the policy analysis tool then back to system trust tool and call to `system_trust` took 0.18 secs

I'll work the changes needed to show the app while the initial `System()` call is loading and caching into the redux branch.

Issue #181, opened 2021-08-05
Pyo3 Enums or ADT for enums in bindings


jw3:
There are now several instances of sending strings over as flags
- trust status
- event subject and object fields
  - access
  - trust
  - mode

There are a finite set of options for each of these, need to figure out the proper way to model a Rust enum into Python.

PR #246, opened 2021-09-20
Trust Analysis

### LINKED ISSUE #235: Roll actual trust status into the analysis bindings
Currently the status is hard coded to populate the GUI.

jw3:
Resolves trust status and provides through bindings to indicate which trust source the event was found in.

Also add some supporting capabilities
- Defines and Analysis type that is derived from various inputs and passed through bindings
- Adds a load function to tdb to support testing trust scenarios that use ancillary
- Example event generator 

closes #235

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -271,6 +271,7 @@ dependencies = [
  "fapolicy-util",
  "lmdb-rkv",
  "nom",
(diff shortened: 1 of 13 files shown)


jw3:
Needs #181

PR #247, opened 2021-09-20
Add dbus agent and pkexec DISPLAY and XAUTHORITY options

tparchambault:


tparchambault:
This appears to work fine on FC34. Technically RHEL8 is also using the Gnome3 desktop and should be Ok but will perform RHEL8 specific testing.

tparchambault:
This modification essentially provides full superuser permissions to the fapolicy-analyzer. 

If my understanding of the polkit design and implementation is in the ballpark, if we want to increase the granularity of permissions, we will have to modify the source to request authority prior to performing an action that requires heightened permissions. Depending on the action different rules can be applied based on the uid, gid and the resource via the action/policy config file. 

This first integration iteration is using pkexec which provides full su power after authentication and does not require the action/policy config file to obtain this capability.

jw3:
This looks fun.  I will review tomorrow.

jw3 (changes requested):
Can you add the necessary components to the playbook so this can be demonstrated from vagrant?

jw3:
On ubuntu it works.  From desktop I get the GUI authorization screen, from console I get a prompt to authenticate.

tparchambault:
> On ubuntu it works. From desktop I get the GUI authorization screen, from console I get a prompt to authenticate.

Which is correct..
polkit-gnome? The package?

jw3:
Yep, add `polkit-gnome` to the playbook.

tparchambault:
I'm surprised it already not installed. In any event remote sessions that start fapolicy-analyzer will still get the text based authentication prompting. (afaik)

jw3:
In my test it said there was no authentication agent available and did not prompt for anything.

tparchambault:
Ok great. Thx for the clarification. This was your Desktop trial or via ssh? I'm trying remember if I had to explicitly add a package to get pkttyagent installed too? If I was back East, I'd just look at history...

tparchambault:
I'll have to probably work with ryan daly too to verify it on the RHEL8 VMs, unless you/Davin have a way for us to used RHEL images in the vagrant set-up.
 The CTC RHEL VMs have a root pw set so polkit requests the superuser pw, which we don't have. To get su authentication on the CTC RHEL VMs one can 'sudo su' which is linked to the RSA authentication process.

tparchambault:
The requested polkit-gnome yum install was added to the system/task/main.yml playbook, but w/o testing. I currently don't have Vagrant set up in my current work env. If you would be so kind. Thx.

jw3:
We still need to authenticate through pkexec, adding a password for root is one way, is that the only way?

PR #249, opened 2021-09-21
Tools update

jw3:
## Clean up the tools crate by
- merging the trust db tools together
- moving the rule builder to examples

## Adds to the trust db tools functions
- additional init from dpkg
- check existing trust against filesystem
- add/delete operations
- rudimentary search
- adds rayon at pinch points

## Also
- Splits the trust read into a checked and unchecked version

## The util

```
#! target/debug/tdb --help
Trust DB Utils v0.1

USAGE:
    tdb [FLAGS] [DBDIR] <SUBCOMMAND>

ARGS:
    <DBDIR>    Specify the fapolicyd database directory Defaults to the value specified in the
               users XDG configuration

FLAGS:
    -h, --help       Print help information
    -v, --verbose    Verbose output
    -V, --version    Print version information

SUBCOMMANDS:
    add       Add a file to the trust database
    check     Check trust status
    clear     Clear the trust database of all entries
    count     Count the number of trust entries
    del       Remove a file from the trust database
    dump      Dump all trust entries to file or stdout
    help      Print this message or the help of the given subcommand(s)
    init      Initialize the trust database from system trust
    search    Search for a trust entry
```

This is laying groundwork capability that leads to support for benchmarking performance of trust machinery.


jw3:
Building a fapolicyd database on Ubuntu can be done with

`tdb init --dpkg`

crates/tools/trust_db_util.rs
```diff
@@ -0,0 +1,352 @@
...
+        }
+    }
+    tx.commit()?;
+
+    let duration = t.elapsed().expect("timer failure");
+    println!(
+        "initialized db with {} entries in {} seconds",
+        sys.len(),
+        duration.as_secs()
+    );
```
scholarsmate, line 217: This is great that we're benchmarking these operations, but I think these are more appropriate to do in verbose mode instead of all the time.  If this tool were to be run as a cron job for instance, if something gets emitted, it generates emails and things.  Most utilities emit nothing and exit with a zero status code on success, unless run in verbose mode.

crates/tools/trust_db_util.rs
```diff
@@ -0,0 +1,352 @@
...
+        clear(ClearOpts {}, cfg, env)?;
+    }
+
+    let db = env.create_db(Some(TRUST_DB_NAME), DatabaseFlags::DUP_SORT)?;
+
+    if opts.empty {
+        return Ok(());
+    }
+
+    let t = SystemTime::now();
```
scholarsmate, line 195: In C++ you can put a timer object on the stack in the same scope as the code being timed.  When the object gets popped off the stack, timing information is emitted in in the destructor.  This provides an elegant standard solution to timing code using minimal lines.  See https://www.boost.org/doc/libs/1_53_0/libs/timer/doc/cpu_timers.html for reference.

crates/tools/trust_db_util.rs
```diff
@@ -0,0 +1,352 @@
...
+    #[clap(long)]
+    empty: bool,
+
+    /// overwrite existing database
+    #[clap(long)]
+    force: bool,
+
+    /// use dpkg to generate db
+    #[clap(long)]
+    dpkg: bool,
```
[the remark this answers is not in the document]
jw3: If there were more than two possible states right now it would have been.  With an enum you will still have RPM as the default and pass a flag to override to dpkg, same as it is now.

crates/tools/trust_db_util.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    match all_opts.cmd {
+        Clear(opts) => clear(opts, &sys_conf, &env),
+        Init(opts) => init(opts, &sys_conf, &env),
+        Add(opts) => add(opts, &sys_conf, &env),
+        Del(opts) => del(opts, &sys_conf, &env),
+        Dump(opts) => dump(opts, &sys_conf),
+        Search(opts) => find(opts, &sys_conf, &env),
+        Check(opts) => check(opts, &sys_conf),
+    }
```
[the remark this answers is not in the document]
jw3: Not matching is impossible.

crates/tools/trust_db_util.rs
jw3: It is the same scope, we are timing both the load and the commit.

crates/tools/trust_db_util.rs
```diff
@@ -0,0 +1,352 @@
...
+        None => Path::new(&sys_conf.system.trust_db_path),
+    };
+
+    if !trust_db_path.exists() {
+        panic!("dbdir does not exist")
+    } else if !trust_db_path.is_dir() {
+        panic!("dbdir must be a directory")
+    }
+
+    println!("opening trust db at {}", trust_db_path.to_string_lossy());
```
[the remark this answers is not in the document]
jw3: Yeah, I can add a verbose flag.

crates/tools/trust_db_util.rs
jw3: Added verbose check here

crates/tools/trust_db_util.rs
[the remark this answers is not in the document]
jw3: Added verbose check here

crates/tools/trust_db_util.rs
```diff
@@ -257,9 +270,11 @@ fn find(opts: SearchDbOpts, _: &cfg::All, env: &Environment) -> Result<(), Error
...
     let db = read::load_trust_db(&cfg.system.trust_db_path)?;
     match opts.outfile {
-        None => db
-            .iter()
-            .for_each(|(k, v)| println!("{} {} {}", k, v.trusted.size, v.trusted.hash)),
+        None => {
+            for (k, v) in db.iter() {
+                println!("{} {} {}", k, v.trusted.size, v.trusted.hash)
+            }
+        }
```
scholarsmate: Why was this change needed?

jw3: It wasn't needed, just matching the style of the other case

Issue #252, opened 2021-09-22
Add developer Ansible role


scholarsmate:
As a developer, I want to be able to easily generate a development VM, so that I can quickly spin up a VM that is ready to go for building and testing the fapolicy-analyzer tool kit and fapolicyd.

Being able to generate a development VM will also foster easier cross-platform development and reduce developer on-boarding.

PR #254, opened 2021-09-22
Add develop Ansible role to site demo (#252)

scholarsmate:
This new role installs the project development dependencies, stops fapolicyd (because the rust compiler is not trusted), builds the software, then restarts fapolicyd.


jw3 (approved):
:+1: lgtm

PR #256, opened 2021-09-22
fix: remove duplicate Close button in About window (#255)

### LINKED ISSUE #255: fapolicy-analyzer's About window has 2 "Close" buttons
The "About" window has 2 "Close" buttons where we only need 1.

scholarsmate:
Removes duplicate "Close" button in the "About" window.

Closes #255


dorschs57:
On my RHEL 8 box I now don't have any close button on the about screen, which would make sense because you delete the only button from the ButtonBox widget.  I'm not sure why you guys are seeing double on a Fedora setup.  The only way I have to close the about box now is escaping it.

image

tparchambault:
As I noted in our other chat forum, and for posterity, on my FC34 VBox VM instance, I see only a single close button on the latest master, i.e. I could not duplicate the original symptom.

jw3:
Verified this is resolved on ubuntu

jw3:
@scholarsmate  is this good to go?

dorschs57:
> @scholarsmate is this good to go?

I verified again on RHEL 8 that there are not no close buttons on the dialog like the pic from my comment [above](https://github.com/ctc-oss/fapolicy-analyzer/pull/256#issuecomment-926547907).  I can exit the about dialog with an escape, but no close button.

PR #258, opened 2021-09-23
Convert yellow analysis entries to green

tparchambault:


tparchambault:
If object access attributes are a function of user and subject selections maybe differing colors would be appropriate. Will bring it up in the tech sync meeting.

tparchambault:
After a tech team discussion, the background color change from LIGHT_YELLOW to LIGHT_GREEN was only applied to the 'object' elements.

jw3 (approved):
:+1:

Issue #257, opened 2021-09-23
Make develop Ansible role conditional and not invoked by default


scholarsmate:
Currently the develop role is run by default.  This slows down testing which doesn't require the develop role.  Make this role optional and make it not run by default.

PR #259, opened 2021-09-23
develop role now conditional and not run by default (#257)

scholarsmate:
To enable the develop role, simply run `ENV="dev" vagrant up` which will configure a shared folder with the host (using Virtualbox's shared folder) so code changes are sync'ed bi-directionally, and the Ansible develop role will install build-time dependencies, checkout, and build the code in that shared location.  Since the shared folder is implemented in the Virtualbox host and guest via Virtualbox guest additions, auto update is enabled when using the dev environment so they are sync'ed properly.  I chose to do it this way over `NFS` because it's more portable, doesn't require privileged access on the host, and less stuff to install (we already have the Virtualbox client additions installed).  Before running `vagrant up`, be sure to create a local .shared directory where the `Vagrantfile` is for the shared storage.

Managing the proxy via the environment in all the Ansible tasks that used the network was getting tedious so this change uses the `vagrant-proxyconf` plugin that makes managing the proxy configuration *much simpler* and less error prone (DRY).

We required the `vagrant-vbguest` plugin and this change adds the `vagrant-proxyconf` plugin so now these vagrant plugins are auto installed if they don't already exist.

The shared folder is `scripts/vagrant/fc34/.shared`, so `.shared` and `.vagrant` directories were added to the `.gitignore` file.

The `.gitignore` file is now sorted, which is a personal preference that I hope will be okay with everyone.

Added generation of RHEL8 VM in addition to the FC34 VM.  The RHEL8 VM requires a RedHat Subscription to use.  So to create the VM, just add a couple more environment variables for the RedHat Subscription Manager so it looks like the following:

```
ENV="dev" RHSM_USERNAME="your_email@ctc.com" RHSM_PASSWORD="your-password-here" vagrant up
```

For the test cases to succeed, we'll need to adjust the URL to one for `el8` since right now we just have `fc34` RPMs published.


jw3 (approved):
:+1: I like the proxy changes.

PR #261, opened 2021-09-24
Support for lists in event logs

### LINKED ISSUE #253: Event log parser to support parsing multiple values
### Problem:
While trying to allow a developer to use the rust compiler (rustc), a log gets generated that looks like this:

```
rule=4 dec=deny_syslog perm=execute uid=1000 gid=190,990,1000 pid=37852 exe=/home/vagrant/.cargo/bin/rustc : path=/home/vagrant/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc ftype=application/x-executable trust=0
```

Currently the parser only expects a single numerical value for the `gid`, but fapolicyd emits the primary group as well as ancillary groups which the parser wasn't expecting.  This error blocks the analyzer from loading the file for analysis.

### Steps to reproduce:

1. Install rust:
```
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
```
2. With fapolicyd running, attempt to execute rustc:
```
~/.cargo/bin/rustc --version
```
3. Generate an event log:
```
journalctl -r | grep -e 'rule=\d+' | cut -d: -f4- | awk '{$1=$1};1' > /tmp/rust-event.log
```
4. Load the event log into the fapolicy-analyzer and on the command line you'll see something like this:
```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Error(Error { input: ",990,1000 pid=37852 exe=/home/vagrant/.cargo/bin/rustc : path=/home/vagrant/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/rustc ftype=application/x-executable trust=0", code: Space })', crates/analyzer/src/log.rs:77:38
```

jw3:
closes #253

### DIFF
--- a/crates/analyzer/src/log.rs
+++ b/crates/analyzer/src/log.rs
@@ -9,6 +9,7 @@ use nom::character::complete::{digit1, space1};
 use nom::sequence::{preceded, terminated};
 
 use crate::rules::*;
+use nom::multi::separated_list1;
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Event {
@@ -16,7 +17,7 @@ pub struct Event {
     pub dec: Decision,
     pub perm: Permission,
     pub uid: i32,
-    pub gid: i32,
+    pub gid: Vec<i32>,
     pub pid: i32,
     pub subj: Subject,
     pub obj: Object,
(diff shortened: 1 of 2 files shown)


jw3:
One lingering issue in here that the bindings api was not updated to support the gid list, instead I am just taking the 0th gid and using that as the value returned to the py side.

This may or may not get solved in this PR.  If not we will address separately.

scholarsmate:
The fapolicy_analyzer no longer crashes when I read an event log file with a comma-separated list of GIDs.  I tested this on FC34 and RHEL8.

PR #270, opened 2021-09-28
generate a deny event for dave the developer

scholarsmate:
Creates a deny event for "dave" the developer who is trying to use a rust compiler that was not installed by a trusted package repository.  The generated event log will look like this:

```{shell}
[vagrant@rhel8 ~]$ grep rustc events-1.txt
1:rule=4 dec=deny_syslog perm=execute uid=1004 gid=100,1002 pid=40358 exe=/usr/bin/bash : path=/home/dave/.cargo/bin/rustc ftype=application/x-executable trust=0
[vagrant@rhel8 ~]$
```

### DIFF
--- a/scripts/demo/develop/tasks/main.yml
+++ b/scripts/demo/develop/tasks/main.yml
@@ -1,6 +1,6 @@
 # install development requirements, and create a build
 ---
-- name: install development requirements common to the RedHat 8 family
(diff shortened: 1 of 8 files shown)


tparchambault (approved):
Visual review only. I do not have an available vagrant testbed.

PR #271, opened 2021-09-28
RHEL 8 build and CI

### LINKED ISSUE #260: Publish RHEL8 RPMs
Right now we only publish Fedora 34 RPMs, but it would be good if we also publish RHEL8 RPMs (and any other targeted platforms) too.  The challenge here is that RHEL instances require RedHat subscriptions (RHSM) that use credentials (secrets) that need to be managed and protected by CI/CD for building of the RHEL RPMs.  How do _the good people_ do this?

jw3:
Adds support into the RPM build for RHEL 8 through UBI 8

closes #260


scripts/rpm/fapolicy-analyzer.spec.j2
```diff
@@ -10,15 +10,16 @@ URL:            https://github.com/ctc-oss/fapolicy-analyzer
...
 
 BuildArch:      x86_64
-BuildRequires:  python3 >= 3.9, python3-devel, python3-pip, python-wheel, dbus-devel, rust-packaging
-{% if (fedora_ver | int) >= 34 %}BuildRequires:  python-setuptools-rust{% endif %}
-Requires:       python3 >= 3.9, gtk3, python3-gobject, python3-events, python3-configargparse
-Requires:       dbus-libs
+Requires:       {{ pyrpm }} >= 3.9
+BuildRequires:  {{ pyrpm }} >= 3.9, {{ pyrpm }}-rpm-macros, {{ pyrpm }}-devel, {{ pyrpm }}-pip, {{ pyrpm }}-wheel
+BuildRequires:  dbus-devel, cargo
+Requires:       gtk3, dbus-libs
```
scholarsmate, line 16: Is it possible to combine the Requires: lines together, and the BuildRequires: lines together?

scholarsmate (commented):
LGTM, but I haven't tested this out yet.

PR #277, opened 2021-09-29
User initiated fapd backup

tparchambault:
Add checkbox to ConfirmInfo dialog to initiate an explicit operation to archive fapolicyd state prior to deploying changes. The user is prompted for the path and filename of the archive.


tparchambault:
Added a new hbox w/a UI checkbox in the area above the OK, Cancel button box and below the pending file list in the ConfirmInfo dialog. If the checkbox is set, a FileChooser dialog is displayed, prompting the user to save the current state (database and config) of the fapolicyd daemon.

fapolicy_analyzer/tests/test_fapd_dbase.py
```diff
@@ -1,13 +1,98 @@
...
+    assert count_fapd_archives() == 1
+
+    # Add another giBackupFileMaxCount+1 of files. Verify max count limit
+    for i in range(giBackupFileMaxCount + 1):
+        assert fapd_dbase_snapshot()
+    assert count_fapd_archives() == giBackupFileMaxCount
+
+    # Clean up: Delete and verify all archives have been removed.
+    delete_fapd_archives()
+    assert count_fapd_archives() == 0
```
dorschs57, line 83: It seems like it would be easier to mock the `subprocess.Popen` here and then assert on the calls to it instead of writing all these helper methods to interact with the file system.

fapolicy_analyzer/tests/test_fapd_dbase.py
```diff
@@ -1,13 +1,98 @@
...
+    pass
+    # strListFile = xdg_config_dir_prefix("fapolicy_backup_manifest.txt")
+    # os.remove(strListFile)
+
+
+@pytest.fixture
+def manifest():
+    strListFile = create_populate_manifest()
+    yield strListFile
+    delete_manifest()
```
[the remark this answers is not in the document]
dorschs57: Also, what is creating this manifest file doing for you?  I don't see where you are specifically testing for anything from it, and the `fapd_dbase_snapshot` method had fallback code if the manifest file doesn't exist.

fapolicy_analyzer/tests/test_fapd_dbase.py
```diff
@@ -1,13 +1,98 @@
...
+        assert fapd_dbase_snapshot()
+    assert count_fapd_archives() == giBackupFileMaxCount
+
+    # Clean up: Delete and verify all archives have been removed.
+    delete_fapd_archives()
+    assert count_fapd_archives() == 0
+
+
+def test_fapd_dbase_cleanup_snapshots_no_argument():
+    assert not fapd_dbase_cleanup_snapshots(None)
```
dorschs57, line 87: Where are you actually testing the `fapd_dbase_cleanup_snapshots` method works with valid arguments?

PR #306, opened 2021-10-19
Add child fd monitoring delay between pkttyagent and pkexec

tparchambault:
Added a conditional delay between the pkttyagent and pkexec calls. Pkttyagent occasionally doesn't complete its dbus registration prior to pkexec requesting authentication services from pkttyagent. 

See the pkttyagent manpage for the specifics of the --notify-fd option. Note that pkttyagent does not close the parent's fd only its own, i.e. the child's copy of the fd.


bin/fapolicy-analyzer
```diff
@@ -2,22 +2,35 @@
 
 # Initialize globals
 pidAgent=0
+strCallbackFile=/tmp/fileCallback.txt
```
jw3, line 5: Let's use a more descriptive name on the temp file here

jw3 (approved):
Success in vagrant :+1:

bin/fapolicy-analyzer
tparchambault: Modified to "/tmp/fapolicy-analyzer-dbus-callback.tmp"

Issue #319, opened 2021-11-10
UI console errors and warnings when switching between tools


dorschs57:
I've been seeing the errors and `GTK-Warnings` below when switching back and forth between the Trust Database Admin tool and the Policy Event Analysis tool. This doesn't always happen.  I believe they are related to navigating away from the Trust Database Admin tool before the System Trust Database tab is fully populated.  It seems like even though we are calling `destroy` on the widget some reference is hanging around waiting for the results for the redux call to `system` to come back.

![image](https://user-images.githubusercontent.com/17859684/141186179-46fc6ecd-6bb9-4191-b73a-fa1273152175.png)

![image](https://user-images.githubusercontent.com/17859684/141186239-d1e618cf-288d-4f42-9ecc-91ad3336893d.png)

To reproduce:
1. Start the application fresh
2. Quickly switch to the Policy Event Analysis tool by going to Action->Analyze from Audit Log and loading a file, before the System Trust Database tab has time to fully load.

PR #320, opened 2021-11-10
Fixed bug with reloading the Policy Event Analysis Tool with a new file

### LINKED ISSUE #315: File Access Policy Analyzer tool does not reload properly
If you are in the File Access Policy Analyzer tool and then select a new log to analyze via the Actions -> Analyze From Audit Log menu item, the tool is not cleared and reloaded with the new data.  If you are in the tool, return to the Trust Database Administration tool, and then select the new file to analyze, when the File Access Policy Analyzer tool reloads it does show the correct data.

dorschs57:
I had to add a 'loading' value to the assorted states that are populated via redux so the value can be checked during the system feature events.

**To Test:**
1. goto Action->Analyze from Audit Log
2. select an event file to load
3. without navigating way from the tool, goto Action->Analyze from Audit Log again
4. select a different event file to load
5. verify the tool is update with the correct data from the new file

NOTE: you might see some errors and warning in the console, but the tool should still load and reload properly.  I believe these errors are related to #319

closes #315

### DIFF
--- a/fapolicy_analyzer/tests/reducers/test_ancillary_trust_reducer.py
+++ b/fapolicy_analyzer/tests/reducers/test_ancillary_trust_reducer.py
@@ -9,41 +9,50 @@ from ui.reducers.ancillary_trust_reducer import (
     handle_error_ancillary_trust,
     handle_error_deploying_ancillary_trust,
     handle_received_ancillary_trust,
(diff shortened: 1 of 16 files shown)


jw3 (approved):
reloads look good

Issue #101, opened 2021-04-21
Config file versioning


jw3:
If the structure of the xdg config file changes in an incompatible way the load will panic

How to manage this?  Optional values, or versioning, not sure.  The priority of this issue depends on how stable that config becomes.

```
thread '<unnamed>' panicked at 'unable to load xdg configuration: BadTomlData(Error { inner: ErrorInner { kind: Custom, line: Some(0), col: 0, at: Some(0), message: "missing field `application`", key: [] } })', /home/addorschs/fapolicy-analyzer/src/cfg.rs:18:35
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Traceback (most recent call last):
  File "/home/addorschs/fapolicy-analyzer/python/fapolicy_analyzer/ui/main_window.py", line 54, in on_start
    page = DatabaseAdminPage().get_content()
  File "/home/addorschs/fapolicy-analyzer/python/fapolicy_analyzer/ui/database_admin_page.py", line 14, in __init__
    SystemTrustDatabaseAdmin().get_content(),
  File "/home/addorschs/fapolicy-analyzer/python/fapolicy_analyzer/ui/system_trust_database_admin.py", line 12, in __init__
    self.system = System()
pyo3_runtime.PanicException: unable to load xdg configuration: BadTomlData(Error { inner: ErrorInner { kind: Custom, line: Some(0), col: 0, at: Some(0), message: "missing field `application`", key: [] } })
```

jw3:
Workaround is to delete the xdg config dir and it will get recreated with a valid version.

`rm -rf $HOME/.config/fapolicy_analyzer`

PR #322, opened 2021-11-15
Rules DB

jw3:
Load and provide lookup of rules.

This is going to break the app when loading the config file #101.


jw3:
> This is going to break the app when loading the config file #101.

Error handling was improved around the config loading, instead of panic we get a semi-understandable message.

> RuntimeError: ConfigError(BadTomlData(Error { inner: ErrorInner { kind: Custom, line: Some(5), col: 0, at: Some(138), message: "missing field `rules_file_path`", key: ["system"] } }))

PR #323, opened 2021-11-16
Provide fapolicyd status and control

tparchambault:
fapolicyd status/control through redux. Daemon polling needs to still be migrated into the redux layer.


fapolicy_analyzer/ui/main_window.py
```diff
@@ -107,6 +125,17 @@ def __set_trustDbMenu_sensitive(self, sensitive):
...
+    def __update_fapd_status(self, bStatus: bool):
+        logging.debug(f"__update_fapd_status({bStatus})")
+        if bStatus:
+            self.fapdStatus.set_from_stock(stock_id="gtk-yes", size=4)
+        else:
+            self.fapdStatus.set_from_stock(stock_id="gtk-no", size=4)
+
+    def on_update_daemon_status(self, status: bool):
+        logging.debug(f"on_update_daemon_status({status})")
+        GLib.idle_add(self.__update_fapd_status, status[1])
```
dorschs57, line 137: You shouldn't have to `GLib` this.  The event should be firing on the UI thread.

fapolicy_analyzer/ui/main_window.py
```diff
@@ -275,3 +304,25 @@ def on_analyzeMenu_activate(self, menuitem, *args):
     def on_trustDbMenu_activate(self, menuitem, *args):
         self.__pack_main_content(router(ANALYZER_SELECTION.TRUST_DATABASE_ADMIN))
         self.__set_trustDbMenu_sensitive(False)
+
+    def on_fapaPreferencesMenu_activate(self, menuitem, data=None):
+        logging.debug("on_fapaPreferencesMenu_activate() invoked.")
+
+    def on_fapdPreferencesMenu_activate(self, menuitem, data=None):
+        logging.debug("on_fapdPreferencesMenu_activate() invoked.")
```
dorschs57, line 312: Are these tied to something?

fapolicy_analyzer/ui/features/daemon_feature.py
```diff
@@ -0,0 +1,179 @@
...
+    def _daemon_start(action: Action) -> Action:
+        logging.debug("_daemon_start(action: Action) -> Action")
+        start_fapolicyd()
+        return received_daemon_start()
+
+    def _daemon_status(action: Action) -> Action:
+        logging.debug(f"_daemon_status(action: {action}) -> Action")
+        status = is_fapolicyd_active()
+        logging.debug(f"_daemon_status::Fapolicyd status: {status}")
+        _fapd_status = status
```
dorschs57, line 117: I don't see a reason to assign a separate variable here just to use for logging.  I'd suggest: 
```suggestion
        _fapd_status = is_fapolicyd_active()
        logging.debug(f"_daemon_status::Fapolicyd status: {_fapd_status}")
```
Same for the `_daemon_status_update` method below.

fapolicy_analyzer/ui/features/daemon_feature.py
```diff
@@ -0,0 +1,179 @@
...
+    provided, a new Daemon object will be initialized.  Used for testing
+    purposes only.
+    """
+    def _init_daemon() -> Action:
+        logging.debug("_init_daemon() -> Action")
+
+        def acquire_daemon():
+            logging.debug("acquire_daemon()")
+            try:
+                daemon = Handle("fapolicyd")
```
dorschs57, line 69: Where does this guy ever get used?  Do you need to initialize a `Handle` before you can use any of the other ' is_fapolicyd_active, start_fapolicyd, stop_fapolicyd' methods?

fapolicy_analyzer/ui/actions.py
```diff
@@ -41,6 +42,29 @@
...
+
+REQUEST_DAEMON_STOP = "REQUEST_DAEMON_STOP"
+RECEIVED_DAEMON_STOP = "RECEIVED_DAEMON_STOP"
+ERROR_DAEMON_STOP = "ERROR_DAEMON_STOP"
+
+REQUEST_DAEMON_STATUS = "REQUEST_DAEMON_STATUS"
+RECEIVED_DAEMON_STATUS = "RECEIVED_DAEMON_STATUS"
+ERROR_DAEMON_STATUS = "ERROR_DAEMON_STATUS"
+
+REQUEST_DAEMON_STATUS_UPDATE = "REQUEST_DAEMON_STATUS_UPDATE"
```
dorschs57, line 60: What is the difference between `REQUEST_DAEMON_STATUS` and `REQUEST_DAEMON_STATUS_UPDATE` actions?

fapolicy_analyzer/ui/actions.py
```diff
@@ -168,3 +192,80 @@ def init_system() -> Action:
...
+def error_daemon_status(error: str) -> Action:
+    return _create_action(ERROR_DAEMON_STATUS, error)
+
+
+def request_daemon_status_update():
+    logging.debug("request_daemon_status_update()")
+    return _create_action(REQUEST_DAEMON_STATUS_UPDATE)
+
+
+def received_daemon_status_update(bStatus: bool):
```
dorschs57, line 236: As far as I can tell this guy is never used.

fapolicy_analyzer/ui/actions.py
dorschs57: To me it looks like you can consolidate the `STATUS` and `STATUS_UPDATE` sets of actions into a single set.  Its ok for multiple places to use the same actions, so you can dispatch the same actions from the `main_window` menu item and from your checker background thread.

dorschs57 (changes requested):

fapolicy_analyzer/ui/main_window.py
tparchambault: No. I pulled the menu items although I was originally considering having a dialog to set the daemon in permissive mode for instance, or configuring the analyzer's color settings. As menu items were removed from the original prototype, removing associated callbacks were in the plans before moving off of Draft status. The priority was/is Redux.

fapolicy_analyzer/ui/features/daemon_feature.py
tparchambault: Just part of the debugging process/noise. I expect to remove a lot of the logging msgs in the reducer/feature modules.

fapolicy_analyzer/ui/features/daemon_feature.py
tparchambault: The Handle does not currently get used.

fapolicy_analyzer/ui/actions.py
tparchambault: We can remove REQUEST_DAEMON_STATUS which is a distinct request invoked from the System menu. My original idea was acquire other fapolicyd info e.g. version, build date, etc, and run status and present it in a dialog when explicitly requested manually via the System menu. The REQUEST_DAEMON_STATUS_UPDATE was associated with the timed polling loop in the feature code. I'll revisit that.

fapolicy_analyzer/ui/actions.py
tparchambault: It was used but did not play well with the manual Start/Stop iirc.It will be removed.

tparchambault:
Thx for the Redux review @dorschs57  As we chatted about, the explicit requests to start/stop the daemon will be moved out of Redux since they only indirectly affect the app state. The associated code will be removed from Redux in addition to the code associated with the above comments.

fapolicy_analyzer/ui/main_window.py
tparchambault: Aspirational code removed.

fapolicy_analyzer/ui/main_window.py
tparchambault: Done.

fapolicy_analyzer/ui/features/daemon_feature.py
tparchambault: Cleaned up.

fapolicy_analyzer/ui/features/daemon_feature.py
tparchambault: The handle/reference is now used for all interfacing to the backend for fapd status/control.

fapolicy_analyzer/ui/actions.py
tparchambault: Removed REQUEST_DAEMON_STATUS and associated variants, RECEIVED, and ERROR.

fapolicy_analyzer/ui/actions.py
tparchambault: Removed code associated with REQUEST_DAEMON_STATUS_UPDATE action.

fapolicy_analyzer/tests/reducers/test_daemon_reducer.py
```diff
@@ -0,0 +1,39 @@
...
+
+@pytest.fixture
+def initial_state():
+    return DaemonState(error=None, status=False, handle=None)
+
+
+def test_handle_received_daemon_start(initial_state):
+    mockDaemon = MagicMock()
+    ret = handle_received_daemon_start(initial_state, mockDaemon)
+    print(ret)
```
dorschs57, line 22: Remove or convert to a `log.debug` statement.

fapolicy_analyzer/tests/reducers/test_daemon_reducer.py
```diff
@@ -0,0 +1,39 @@
...
+def test_handle_received_daemon_reload():
+    pass
+
+
+def test_handle_received_daemon_status():
+    pass
+
+
+def test_handle_received_daemon_status_update():
+    pass
```
dorschs57, line 39: Remove these before the final merge unless you are going to add real tests to them.

fapolicy_analyzer/ui/strings.py
```diff
@@ -6,6 +6,7 @@
 SYSTEM_INITIALIZATION_ERROR = _("Error initializing System")
 ANCILLARY_TRUST_LOAD_ERROR = _("Error loading Ancillary Trust")
 SYSTEM_TRUST_LOAD_ERROR = _("Error loading System Trust")
+DAEMON_INITIALIZATION_ERROR = _("Error initializing communications with daemon")
```
dorschs57, line 9: You need to run `pipenv run extract_messages` to update the pot file.

fapolicy_analyzer/ui/actions.py
```diff
@@ -62,6 +81,12 @@ class Notification(NamedTuple):
     type: NotificationType
 
 
+class DaemonState(NamedTuple):
+    error: str
+    status: bool
+    handle: Handle
```
dorschs57, line 87: I don't see anywhere this `handle` is being used.  Also it shouldn't really be exposed in the application state.  If something what's to interact with the ref being held by the redux feature, they should be doing through some action.

fapolicy_analyzer/ui/actions.py
```diff
@@ -41,6 +42,24 @@
...
+REQUEST_DAEMON_STOP = "REQUEST_DAEMON_STOP"
+RECEIVED_DAEMON_STOP = "RECEIVED_DAEMON_STOP"
+ERROR_DAEMON_STOP = "ERROR_DAEMON_STOP"
+
+RECEIVED_DAEMON_STATUS_UPDATE = "RECEIVED_DAEMON_STATUS_UPDATE"
+ERROR_DAEMON_STATUS_UPDATE = "ERROR_DAEMON_STATUS_UPDATE"
+
+REQUEST_DAEMON_RELOAD = "REQUEST_DAEMON_RELOAD"
+RECEIVED_DAEMON_RELOAD = "RECEIVED_DAEMON_RELOAD"
+ERROR_DAEMON_RELOAD = "ERROR_DAEMON_RELOAD"
```
dorschs57, line 61: It doesn't look like any of your `ERROR` actions are ever handled by a reducer that would update the state with the `error` value.  Have you tested error handling?

dorschs57 (changes requested):

fapolicy_analyzer/ui/features/daemon_feature.py
```diff
@@ -0,0 +1,162 @@
...
+                handle=_fapd_ref,
+                error=None
+            )))
+
+        if daemon:
+            finish(daemon)
+        else:
+            acquire_daemon()
+
+        if "NO_DAEMON_MONITORING" not in os.environ:
```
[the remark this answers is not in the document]
tparchambault: It can be set to anything to avoid starting the monitoring thread.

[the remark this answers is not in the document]
tparchambault: It's there only to not start monitoring when unit-testing.

fapolicy_analyzer/ui/features/daemon_feature.py
```diff
@@ -0,0 +1,162 @@
...
+            start_daemon_monitor()
+
+        return init_daemon()
+
+    def _daemon_reload(action: Action) -> Action:
+        logging.debug("_daemon_reload(action: Action) -> Action")
+        return received_daemon_reload()
```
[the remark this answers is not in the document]
tparchambault: I wanted to talk with the team wrt user workflow. If new rules are added manually through the rules.d directory for instance do we want a menu item to reread/reload the input config files.

fapolicy_analyzer/tests/reducers/test_daemon_reducer.py
tparchambault: Sorry I thought I deleted this file from the branch. Will try to remove it again.

fapolicy_analyzer/ui/actions.py
tparchambault: Ok. I was thinking that if it's not set i.e. None, then the ui layer could infer it's not installed and dim out all backend operation elements. But your point about not exposing it to the UI layer is a good one, Should we have just a boolean?

fapolicy_analyzer/ui/actions.py
tparchambault: They were there for consistency with the other actions. Right now there is no error handling.

dorschs57: Ok, I'd like to see error handling added before merging this.

fapolicy_analyzer/ui/actions.py
dorschs57: What if you make the `status` 3 states `(None, True, False)` where `None` would indicate that its not installed or the status cannot be pulled.

fapolicy_analyzer/ui/features/daemon_feature.py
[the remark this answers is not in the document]
dorschs57: Ok, but I'd still expect if I set `NO_DAEMON_MONITORING=FALSE` that the system would still turn on daemon monitoring.   Maybe try
```suggestion
        if os.getenv("NO_DAEMON_MONITORING", "false").lower() == "true":
```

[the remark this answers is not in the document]
tparchambault: I like it. Done.

fapolicy_analyzer/tests/reducers/test_daemon_reducer.py
tparchambault: This file should have been removed.

fapolicy_analyzer/ui/strings.py
tparchambault: Done

fapolicy_analyzer/tests/reducers/test_daemon_reducer.py
tparchambault: (Hopefully) Done.

fapolicy_analyzer/ui/features/daemon_feature.py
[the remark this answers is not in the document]
tparchambault: The dynamic HANDLE doesn't support reload currently. Removed all code associated w/reloading the fapd daemon.

fapolicy_analyzer/ui/actions.py
tparchambault: Currently revisiting DaemonState...

tparchambault:
The only thing I've added is code that doesn't start the monitoring thread if fapolicyd is not installed and dispatch an error. I do not currently change the app state based on that error, given that this 'error' may actually be an acceptable run state based on our on-going discussion re: user workflow and use-cases.

fapolicy_analyzer/ui/actions.py
tparchambault: Error state actions are generated in the exception block in the daemon_feature.py code. I did not add associated reducer code because fapd return errors/exceptions are still TBD and did not know what the next state of the app should be, i.e. what do we want to do if Start or Stop fail? Notification pop-up? or termination? There are associated event dispatches with the errors but no changes to the app state occur.

fapolicy_analyzer/ui/actions.py
tparchambault: Made status an enum TRUE, FALSE, UNKNOWN.

tparchambault:
See PR #369 for the simpler, alternative implementation.

PR #324, opened 2021-11-16
Updating gui to support detail scroll and avoid resizing

egbicker:
Added a requested size to the details panel and made the detail panels homogenous in height and width. Added the ability to scroll within the details panel if the information does not fit within the requested size. Detail panels can be resized individually by dragging the pane boundary but this breaks homogeneousness.


fapolicy_analyzer/glade/policy_rules_admin_page.glade
```diff
@@ -1,145 +1,211 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.38.2 -->
+<!-- Generated with glade 3.22.2 -->
 <interface>
-  <requires lib="gtk+" version="3.22"/>
-  <object class="GtkBox" id="policyRulesAdminPage">
+  <requires lib="gtk+" version="3.20"/>
```
dorschs57, line 4: Is there a reason the GTK version was downgraded to 3.20?

dorschs57:
Also looks like you have the dash vs underscore problem.  You must be using an older version of glade?

egbicker:
I must be. I don't think I changed anything in my setup so I'll have to go take a look. Thanks for spotting that.

fapolicy_analyzer/glade/policy_rules_admin_page.glade
egbicker: Nope. New commits should have this all fixed

fapolicy_analyzer/ui/policy_rules_admin_page.py
```diff
@@ -369,7 +369,6 @@ def __switch_change_handlers(self, lst, event, handler):
             handlers.__iadd__(handler)
 
         def set_as_primary(self):
-            self.__panel.get_parent().reorder_child(self.__panel, 0)
```
dorschs57, line 372: Removing this seems to have broken the panel switcher.  I can no longer switch the 1st and 2nd panels around using the double arrow button.

fapolicy_analyzer/glade/trust_file_details.glade
```diff
@@ -2,18 +2,21 @@
 <!-- Generated with glade 3.38.2 -->
 <interface>
   <requires lib="gtk+" version="3.22"/>
-  <object class="GtkBox" id="trustFileDetails">
+  <!-- n-columns=3 n-rows=3 -->
```
dorschs57, line 5: In the glade file you seem to still have extra columns that are not being used for anything.  You should change the grid properties to get rid of those.
image

fapolicy_analyzer/glade/policy_rules_admin_page.glade
```diff
@@ -2,14 +2,16 @@
 <!-- Generated with glade 3.38.2 -->
 <interface>
   <requires lib="gtk+" version="3.22"/>
-  <object class="GtkBox" id="policyRulesAdminPage">
+  <!-- n-columns=3 n-rows=3 -->
```
dorschs57, line 5: In glade you have extra rows that don't seem to be used. Update the properties to remove those.
image

dorschs57:
With this new layout on the Policy Analysis Admin page, can we force the details sections at the bottom of the columns to say in sync with each other?  For example if I resize one of them could the other 2 resize automatically to stay in sync.  I think that would look better.
image

What you have here is good enough for this ticket.  If this change is complicated I'd say add a new ticket for it.

fapolicy_analyzer/glade/trust_file_details.glade
egbicker: Fixed

fapolicy_analyzer/glade/policy_rules_admin_page.glade
egbicker: Fixed by transferring back to a Box object

fapolicy_analyzer/ui/policy_rules_admin_page.py
egbicker: Fixed by reverting to a Box object. Test updated as well.

egbicker:
> With this new layout on the Policy Analysis Admin page, can we force the details sections at the bottom of the columns to say in sync with each other? For example if I resize one of them could the other 2 resize automatically to stay in sync. I think that would look better. image
> 
> What you have here is good enough for this ticket. If this change is complicated I'd say add a new ticket for it.

I'll go ahead and open a new ticket. It's going to be involved to figure out how to make everything play nice on two axes

dorschs57:
> > With this new layout on the Policy Analysis Admin page, can we force the details sections at the bottom of the columns to say in sync with each other? For example if I resize one of them could the other 2 resize automatically to stay in sync. I think that would look better. image
> > What you have here is good enough for this ticket. If this change is complicated I'd say add a new ticket for it.
> 
> I'll go ahead and open a new ticket. It's going to be involved to figure out how to make everything play nice on two axes

sound good.

dorschs57:
Everything test out good now and the glade file are looking better.

PR #327, opened 2021-11-19
Modify wrapper to pass args

tparchambault:
Minor bash change to pass the command line arguments to the wrapped python program.


jw3 (approved):
:+1:

PR #329, opened 2021-11-20
Rollback on unconfirmed deployment

### LINKED ISSUE #13: Guard rule application with timed confirmation
If too restrictive of a rule was put in place it could lock the system, but before the rules are applied the daemon needs bounced.  After the bounce set a timer for X number of seconds (15?) that once expired will shutdown the daemon.

This behavior is commonly seen in video display settings, _"The display has been modified. Do you want to keep these settings?"_

jw3:
Implement rollback in the case of a non-confirmation after deployment, which results in the system rolling back to the previous state.

This is somewhat handy when dealing only in trust, but it becomes very useful when we progress to altering rules.

closes #13


jw3:
@dorschs57 I need to provide the system to roll back to from the UI and don't know the proper redux way to do that.

jw3:
The rpms from the last build here need tested in vagrant.

tparchambault:
Tested on a Fedora34 VM w/an rpm install, and fapolicyd initially not running, but on a second attempt to deploy, fapolicyd was running. The 'deploy' operation resulted in an Error pop-up and Try again message, so did not reach the count-down dialog to verify the ability to System rollback. On restart of the fapolicy-analyzer tool, the files that I attempted to add to the ATDb were visible in the ATDA view and marked as 'Trusted. 

I used the x86_64 rpm artifact associated with the Action > Run 729.

jw3:
Going to get the redux vendoring sorted out and then wrap this up on top of it.

jw3:
This is rebased on redux vendoring now, the rpm from ci should be testable.

tparchambault:
On a clean FC34 system, observed that the deploy/countdown restore user workflow works as expected, however I also observed that fapolicyd is stopped and not restarted. Could the return value of the deploy cause the following ".and_then(|_| start_fapolicyd())" to not to be executed?

tparchambault:
@jw3, the majority of the time rollback works as expected but I'm seeing a lot of unexpected behaviors. I will retest such that I can give definite steps to reproduce the issues. E.g. Failed deployments (as per the status pop-up) when fapolicyd is not running, yet the failed operation is visible in the ATDA view on restart. Conceptually, I'm Ok w/that because direct modification of the database is possible independent of fapolicyd's runstate. Most rollbacks appeared to work, i.e. files were not visible on restart, but occasionally they were. This may be an artifact of my test run having more than one sequential operation. I will repeat testing with only one operation per run to isolate symptoms.

I used the artifacts of the last Rollback on unconfirmed deployment RPM run

tparchambault:
Checkpoint: Out of the box FC34 w/sshd enabled and executing.
- Basic test prep: fapolicyd daemon installation and config, etc
```
$ sudo dnf install emacs
$ sudo dnf install fapolicyd
$ sudo emacs /usr/lib/systemd/system/fapolicyd.service
# Add --permissive option, i.e. "ExecStart=/usr/sbin/fapolicyd --permissive"
$ sudo systemctl enable fapolicyd.service
$ sudo systemctl start fapolicd.service
# After scp'ing and unzipping the rpm zip file to the target's home dir
$ sudo dnf install ././fapolicy-analyzer-0.0.0-1.170.g1b6b9bb.fc34.x86_64.rpm
$ fapolicy-analyzer -v
``` 
The above command sequence was repeated after rolling the testbed VM back to sshd checkpoint.

1. Added /root/anaconda-ks.cfg to the ATDb; Terminated and restarted the fapa tool; anaconda-ks.cfg was correctly in the ATDb.
2. In the same runtime session, added /home/tom/.bash_history; Deployed and then selected 'No' on the countdown dialog; Terminated the tool and immediately checked the status of the daemon w/`sudo systemctl status fapolicyd.service` which stated that the service was inactive (dead) with the following log messages:
```
[toma@fedora ~]$ sudo systemctl status fapolicyd.service 
[sudo] password for toma: 
○ fapolicyd.service - File Access Policy Daemon
     Loaded: loaded (/usr/lib/systemd/system/fapolicyd.service; enabled; vendor>
     Active: inactive (dead) since Tue 2021-12-07 19:55:59 EST; 3min 36s ago
    Process: 31028 ExecStart=/usr/sbin/fapolicyd --permissive (code=exited, sta>
   Main PID: 31029 (code=exited, status=0/SUCCESS)
        CPU: 5.110s

Dec 07 19:55:51 fedora fapolicyd[31029]: Updating database
Dec 07 19:55:51 fedora fapolicyd[31029]: Creating database
Dec 07 19:55:51 fedora fapolicyd[31029]: Loading data from rpmdb backend
Dec 07 19:55:51 fedora fapolicyd[31029]: Loading data from file backend
Dec 07 19:55:51 fedora fapolicyd[31029]: Updated
Dec 07 19:55:58 fedora systemd[1]: Stopping File Access Policy Daemon...
Dec 07 19:55:58 fedora fapolicyd[31029]: shutting down...
Dec 07 19:55:59 fedora systemd[1]: fapolicyd.service: Deactivated successfully.
Dec 07 19:55:59 fedora systemd[1]: Stopped File Access Policy Daemon.
Dec 07 19:55:59 fedora systemd[1]: fapolicyd.service: Consumed 5.110s CPU time.
lines 1-17/17 (END)
```
Restarting the fapolicyd daemon and then the fapa tool:
```
The prior rollback operation appears to have been accomplished correctly; The issue is that the daemon is not restarted or is terminated post rollback.

jw3:
After testing some I think its a race situation causing the deploy to fail.  The dbus calls are blocking, but only to the point of the execution of the systemd operation, not to the point of the systemd operation completion.

That is speculation.  I pushed a change to check this.

tparchambault:
Tested w/latest rpm artifacts: 
6 deploy/rollback operation sequences performed: daemon was restarted 3 or the 6 times.

PR #332, opened 2021-11-22
Redux support from RPM

### LINKED ISSUE #242: Need to integrate redux-py and rx libraries into the RPM build

jw3:
Update RPM installation to support the redux components.

closes #242


tparchambault (approved):

dorschs57 (approved):

jw3:
There was an alternate approach to vendoring dependencies, which was to have a vendored directory at the top level.  While I am leaning towards merging this as-is since it is tested, if the future brings more depdencies that need vendored then it might make sense to make a top level directory that simplifies the vendoring process.  At that time the redux-py stuff could be extracted and moved up.

jw3:
> FAIL Required test coverage of 90.0% not reached. Total coverage: 85.44%

tparchambault:
It was passing the CI pipeline but the branch had gotten stale with respect to master. As I mentioned in the status report, the fapd dbase script is getting zero coverage vs its normal 100% Troubleshooting in progress. Logs report some complaining about not being able to write tar file. We have move to the integrated Tarfile module, but there may be an issue with target directory permissions on the CI test platform. Locally unit-test have ~97% coverage. It's possible the base Ubuntu image may have been updated yielding a slightly different env which we've dealt with in the past.

tparchambault:
Actually all tests passed. Will try and get a more verbose output out of the CI pipeline.

tparchambault:
redux_rpm_ci_unit_tests

The CI run is not exercising the code under ui/reducers

tparchambault:
redux_rpm_local_unit_tests

Local unit testing does not exhibit this issue.

tparchambault:
I am going to guess that is because I have redux-py in my virtual env. Stand-by.

dorschs57:
I don't see any tests under the redux code, so I'm assuming those modules are only picking up coverage from our own tests that use them.  I'm not sure why that doesn't happen is CI though.  We could add the `--ignore` flag on the pytest cli to ignore the redux modules.  This might be a other good reason to do as @jw3 suggested and move the redux code up a level though. The pytest command is collecting anything under the `fapolicy_analyzer/` directory.

dorschs57:
The other odd thing is that @jw3 original PR before I changed the imports to fix the application, ran the test successfully.  It had coverage on those modules under `redux\`. https://github.com/ctc-oss/fapolicy-analyzer/runs/4290922654?check_suite_focus=true

Maybe something with the way I imported redux into the UI modules?

tparchambault:
Running a single unit test: ".../tests/reducers/test_ancillary_trust_reducer.py"

tparchambault:
ModuleNotFoundError_redux_rpm

So much for drop n drag...

tparchambault:
I 'good' local run has no issue with the import and finds the .rust module in the shared object.
`# extension module 'fapolicy_analyzer.rust' loaded from '/home/toma/Development/CTC/fapolicy-analyzer/fapolicy_analyzer/rust.cpython-39-x86_64-linux-gnu.so'`

jw3:
Replaced with #340

PR #333, opened 2021-11-23
Issue 316 trust reconciliation dialog

### LINKED ISSUE #316: Add trust reconciliation screen to Policy Event Analysis tool
Add the ability to reconcile the ancillary trust for a Subject or Object from the Policy Event Analysis tool.  This would look simular to the reconciliation that is available from the Ancillary Trust Database Admin tool.

dorschs57:
Adds a trust reconciliation dialog to the Policy Event Analysis tool.  It can be opened by either double clicking a subject or object, or right clicking it and selecting `Reconcile` from the context menu.  The dialog uses the following rules:

* If the file is trusted in the system DB it is view only
* if the file is untrusted in the system DB or the ancillary DB you can Trust it
* if the file is trusted in the ancillary DB you can Untrust it
* if the file is unknown you can Trust it

These will generate changesets that can then be applied by going to the ADTA tool and deploying. Eventually the deploy will be available from the toolbar we talked about.

closes #316

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -31,6 +31,7 @@ events = "0.4"
 argparse = "1.4.0"
 redux-py = "*"
 rx = "*"
+more-itertools = "*"
 
 [pipenv]
 allow_prereleases = true
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "e03f57591887611c49cc41d2ffb482c81d64cce8772e373b784b3ffd7052f67d"
+            "sha256": "cccbd35357033246e1bc626e10c676c9c7c64677f34f2abae086c927e4eb3202"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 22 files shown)


tparchambault (approved):
What is your style for ordering the module imports? I'm assuming it's alphabetical, but then I see some that break that model. Or is this a black formatting artifact?

jw3 (changes requested):
Add the more-itertools package to the rpm spec file so we can test the rpm prior to merge.

Otherwise looks good.

jw3:
One note;  if you trust the same file twice it shows up in the review changeset dialog twice.

dorschs57:
> Add the more-itertools package to the rpm spec file so we can test the rpm prior to merge.
> 
> Otherwise looks good.

Added to the spec file.

dorschs57:
> One note; if you trust the same file twice it shows up in the review changeset dialog twice.

I believe this a consequence of the Policy Analysis tool not taking changesets into account when they are applied.  I've generated a ticket for this because its actually a bigger issue that goes back to the Policy Event Analysis Tool, not just this reconciliation dialog.  #339

jw3 (approved):

dorschs57:
Merging this with the clippy errors since it does not make any changes to the rust code.  The clippy errors are being handled by #343.

PR #340, opened 2021-12-02
Vendor redux-py for RPM build

### LINKED ISSUE #242: Need to integrate redux-py and rx libraries into the RPM build

jw3:
An alternative approach to providing `redux-py` to the RPM installations.  The need for this is because there is no official package available for `redux-py` to be installed through RPM dependencies, so we need to provide it in some way that is compatible with the upstream source RPM requirements.

The attempt in #332 is now closed in favor of this one.

In #332 we copied `redux-py` source into our base package, what some languages call shading.  This PR instead vendors the entire `redux-py` project into our repo and installs it using pip as-is for the rpm and dev environments.  

This one ends up much cleaner, as In #332 we had changes in all of our Python source that imported `redux` to change the import to the shaded library path.  In this PR _none_ of our source is touched, the only changes are to the Pipfile, the lock file, and the RPM spec file.

Pip is used during the RPM install to install `redux-py` and we change the pipfile to use the vendored path to the library for dev environments.

Cleaner than expected.

Closes #242


jw3:
Please test your local dev env and the rpm from ci - https://github.com/ctc-oss/fapolicy-analyzer/suites/4561287527/artifacts/122560700

vendor/ReduxPY/setup.py
```diff
@@ -0,0 +1,50 @@
...
+        'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
+        'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
+        "Programming Language :: Python :: 3",
+        'Typing :: Typed'
+    ],
+    install_requires=read_requirements('requirements.txt'),
+    tests_require=read_requirements('test-requirements.txt'),
+    packages=find_packages(exclude=("tests", "tests.*",)),
```
jw3, line 47: This was the only modification of `redux-py`.  It looks like there is a bug in that repo that causes the `tests` package to install with the library.  That same behavior is observable from pypi as well.

tparchambault:
> Please test your local dev env and the rpm from ci - https://github.com/ctc-oss/fapolicy-analyzer/suites/4561287527/artifacts/122560700

Rpm install on an FC34 was good w/no observable issues wrt rx and redux (expected rollback issue was observed). Dev environment testing currently in-progress...

tparchambault:
In a dev environment, the redux module was installed under ~/.local/share/virtualenvs/fapolicy-analyzer-<chksum>/lib/python3.9/site-packages/redux/. If you are Ok w/that install location, then I believe we are good to go.

tparchambault (approved):

dorschs57 (approved):
Tests out fine in my dev environment.

PR #343, opened 2021-12-02
Address rust clippy/linting warnings

tparchambault:
New rust tool release has generated new warning about our codebase that case the CI pipeline the fail.


jw3:
lgtm

PR #344, opened 2021-12-10
Added an application toolbar with deploy button

### LINKED ISSUE #341: Add an application toolbar
Add a toolbar to the application for a place the user can perform common actions like deploying change to fapolicyd from anywhere in the UI.  The toolbar should support the ability for each tool to have its own set of action specific to it added to the toolbar when the user navigates to that tool.

dorschs57:
This is an initial cut of adding a toolbar to the application.  I picture this being built out a bit more with more buttons, and buttons that are specific to each tool.  However for now my goal was to get a Deploy button out there for us from the Policy Event Analysis tool.  I've created the concept of a "operation" to house code that needs to be used for triggering something from multiple places.  In this case I needed to reuse the deployment workflow from both the button on the ADTA tool and the toolbar button.  This should allow use to also create a menu item if we want to to trigger a deployment.

closes #341

### DIFF
--- a/fapolicy_analyzer/glade/main_window.glade
+++ b/fapolicy_analyzer/glade/main_window.glade
@@ -54,7 +54,7 @@
             <property name="can-focus">False</property>
             <property name="orientation">vertical</property>
             <child>
-              <object class="GtkMenuBar">
+              <object class="GtkMenuBar" id="mainMenu">
                 <property name="visible">True</property>
                 <property name="can-focus">False</property>
                 <child>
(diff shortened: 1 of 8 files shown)


tparchambault (approved):
IMO the new deploy widget on the toolbar should be grayed out, i.e. unavailable to generate UI events when the changeset is empty; Similar to how the ATDA rectangular "Deploy" button behaves currently, where it is disabled until a trust change has been added to the queue.

dorschs57:
> IMO the new deploy widget on the toolbar should be grayed out, i.e. unavailable to generate UI events when the changeset is empty; Similar to how the ATDA rectangular "Deploy" button behaves currently, where it is disabled until a trust change has been added to the queue.

That's a good observation.  I meant to do that from the beginning, but forgot about it.  I'll get that wired up before we merge this one.

dorschs57:
> IMO the new deploy widget on the toolbar should be grayed out, i.e. unavailable to generate UI events when the changeset is empty; Similar to how the ATDA rectangular "Deploy" button behaves currently, where it is disabled until a trust change has been added to the queue.

Fixed

PR #345, opened 2021-12-17
Insert boilerplate OSS license

tparchambault:
Developed a general utility to insert/delete license text into source files. Inserted licensing text into our py files. I still need to determine what the comment characters are for the glade xml files.


tparchambault:
Works fine w/rs files.

jw3:
see #346

tparchambault:
Redundant.

PR #352, opened 2021-12-20
Service handle validity check

jw3:
Provide status checking method on a service handle.

### DIFF
--- a/crates/daemon/src/error.rs
+++ b/crates/daemon/src/error.rs
@@ -6,6 +6,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use std::string::FromUtf8Error;
 use thiserror::Error;
 
 // errors that can occur in this crate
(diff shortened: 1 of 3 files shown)


tparchambault:
Merged these changes into my Daemon Control and Status Monitoring branch. The CI generated rpm was installed on a stripped FC34 w/o fapolicyd installed. Daemon Handle was classified valid by the is_valid() call. Here are the logging.debug() messages:
```
DEBUG:root:acquire_daemon()
ERROR:root:Error initializing System
  File "/usr/lib/python3.9/site-packages/fapolicy_analyzer/ui/features/system_feature.py", line 89, in execute_system
    system = System()
RuntimeError: TrustError(LmdbNotFound("/var/lib/fapolicyd"))
-> DEBUG:root:daemon = <svc.Handle object at 0x7f58b6972c90>
-> DEBUG:root:daemon.is_valid() = True
DEBUG:root:daemon_feature::finish(<svc.Handle object at 0x7f58b6972c90>)
```
SD and I are aware that the System() is currently being requested prior to acquiring the daemon handle, and we will switch the order, but that's secondary for this discussion. It would be useful for the Daemon Handle to either be None or is_valid() return false when either fapolicyd is not installed or has never been run such that /var/lib/fapolicyd does not exist, so that the UI/app can grey out certain functionality and notify the user. Alternatively, we could make fapolicyd a dependency in the spec file.

We can talk about this tomorrow in the status meeting. Thx @jw3

jw3:
There was a bug, `is_valid()` should work now.  Expectation is it will throw if fapolicyd unit does not exist, return the status of the service otherwise.

PR #360, opened 2021-12-21
Fa206 multi select

egbicker:


fapolicy_analyzer/ui/ancillary_trust_database_admin.py
```diff
@@ -76,47 +76,48 @@ def delete_trusted_files(self, *files):
             changeset.del_trust(file)
         self.__apply_changeset(changeset)
 
-    def on_trust_selection_changed(self, trust):
-        self.selectedFile = trust.path if trust else None
+    def on_trust_selection_changed(self, trusts):
+        self.selectedFile = trusts if trusts else None
```
dorschs57, line 80:
```suggestion
        self.selectedFiles = [t.path for t in trusts] if trusts else None
```

fapolicy_analyzer/ui/ancillary_trust_database_admin.py
```diff
@@ -76,47 +76,48 @@ def delete_trusted_files(self, *files):
...
-            trustBtn.set_sensitive(not trusted)
-            untrustBtn.set_sensitive(trusted)
-
-            if isinstance(trust, Trust):
-                self.trustFileDetails.set_in_database_view(
-                    f(
-                        _(
-                            """File: {trust.path}
+        for trust in trusts:
+            if trust:
```
dorschs57, line 84: I'm not sure why you need to loop here.  We can only display 1 file at a time in the details, so you should just pull either the first or last one off the list to display to make this more efficient.

fapolicy_analyzer/ui/policy_rules_admin_page.py
```diff
@@ -324,29 +324,33 @@ def on_next_system(self, system):
...
-        uid = data[1] if data else None
-        if uid != self.selectedUser:
-            self.selectedUser = uid
-            self.selectedGroup = None
-            self.__populate_acl_details(uid, acl.getUserDetails)
-            secondaryAction()
+        if data:
+            uids = [datum[1] for datum in data]
+        else:
+            uids = [None]
```
dorschs57, line 330:
```suggestion
         uids = [datum[1] for datum in data] if data else [None]
```

fapolicy_analyzer/ui/policy_rules_admin_page.py
```diff
@@ -324,29 +324,33 @@ def on_next_system(self, system):
...
-        if subject == self.selectedSubject:
+        gids = [datum[1] for datum in data] if data else [None]
+        for gid in gids:
+            if gid != self.selectedGroup:
+                self.selectedGroup = gid
+                self.selectedUser = None
+                self.__populate_acl_details(gid, acl.getGroupDetails)
+                secondaryAction()
+
+    def on_file_selection_changed(self, secondaryAction, data):
```
dorschs57, line 347: I think keeping this as `on_subject_selection_changed` makes it more clear what it is used for.

fapolicy_analyzer/ui/searchable_list.py
```diff
@@ -114,8 +113,11 @@ def get_action_buttons(self):
         return self.get_object("actionButtons").get_children()
 
     def on_view_selection_changed(self, selection):
-        model, treeiter = selection.get_selected()
-        data = model[treeiter] if model and treeiter else []
+        model, treeiter = selection.get_selected_rows()
+        if model and treeiter:
+            data = [model[i] for i in treeiter]
+        else:
+            data = []
```
dorschs57, line 120:
```suggestion
        data = [model[i] for i in treeiter] if model and treeiter else []
```

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -108,8 +108,12 @@ def __color(self, access):
...
 
     def __handle_selection_changed(self, data):
-        fileObj = data[3] if data else None
-        self.file_selection_changed(fileObj)
+        if data:
+            fileObjs = [datum[3] for datum in data]
+            for fileObj in fileObjs:
+                self.file_selection_changed(fileObj)
+        else:
+            self.file_selection_changed(None)
```
dorschs57, line 116:
```suggestion
        fileObjs = [datum[3] for datum in data] if data else None
        self.file_selection_changed(fileObj)
```

fapolicy_analyzer/ui/system_trust_database_admin.py
```diff
@@ -86,10 +86,11 @@ def on_next_system(self, system):
...
             self.trustFileList.load_trust(self._trust)
 
-    def on_trust_selection_changed(self, trust):
-        self.selectedFile = trust
+    def on_trust_selection_changed(self, trusts):
+        self.selectedFile = trusts
         addBtn = self.get_object("addBtn")
-        if trust:
+        if trusts:
+            trust = trusts[-1]
```
dorschs57, line 93: 👍

fapolicy_analyzer/ui/trust_file_list.py
```diff
@@ -68,7 +68,10 @@ def __init__(self, trust_func, markup_func=None, *args):
         self.get_ref().connect("destroy", self.on_destroy)
 
     def __handle_selection_changed(self, data):
-        trust = data[3] if data else None
+        if data:
+            trust = [datum[3] for datum in data]
+        else:
+            trust = None
```
dorschs57, line 74:
```suggestion
        trust = [datum[3] for datum in data] if data else None
```

dorschs57 (changes requested):

dorschs57:
I'm seeing an error `NoneType object is not iterable` when I multiselect or single select a file from the System Trust Database Admin screen and click Add to Ancillary Trust button.

dorschs57:
> I'm seeing an error `NoneType object is not iterable` when I multiselect or single select a file from the System Trust Database Admin screen and click Add to Ancillary Trust button.

I see the same thing on the Ancillary Trust Database Admin screen when I attempt to trust/untrust files.

egbicker:
Fixed the `NoneType object is not iterable` issue and all the code changes you mentioned

fapolicy_analyzer/ui/policy_rules_admin_page.py
```diff
@@ -324,34 +324,35 @@ def on_next_system(self, system):
...
-            self.selectedGroup = None
-            self.__populate_acl_details(uid, acl.getUserDetails)
-            secondaryAction()
+        uids = [datum[1] for datum in data] if data else [None]
+        for uid in uids:
+            if uid != self.selectedUser:
+                self.selectedUser = uid
+                self.selectedGroup = None
+                self.__populate_acl_details(uid, acl.getUserDetails)
+                secondaryAction()
```
dorschs57, line 333: Much like on the ATDA, you are wasting cycles by looping over the list here.  We are only displaying the last selected users data.  In fact allowing multiselect on the users and groups is pointless as it is right now.  The user can only act on the last selected user.  Are there plans to merge the lists from multi users/groups for the secondary action?

fapolicy_analyzer/ui/policy_rules_admin_page.py
```diff
@@ -324,34 +324,35 @@ def on_next_system(self, system):
...
-            self.selectedUser = None
-            self.__populate_acl_details(gid, acl.getGroupDetails)
-            secondaryAction()
+        gids = [datum[1] for datum in data] if data else [None]
+        for gid in gids:
+            if gid != self.selectedGroup:
+                self.selectedGroup = gid
+                self.selectedUser = None
+                self.__populate_acl_details(gid, acl.getGroupDetails)
+                secondaryAction()
```
dorschs57, line 342: See my comment above on the `on_user_selection_changed` method.

fapolicy_analyzer/ui/system_trust_database_admin.py
```diff
@@ -123,4 +124,5 @@ def on_trust_selection_changed(self, trust):
 
     def on_addBtn_clicked(self, *args):
         if self.selectedFile:
-            self.file_added_to_ancillary_trust(self.selectedFile.path)
+            for sfile in self.selectedFile:
+                self.file_added_to_ancillary_trust(sfile.path)
```
dorschs57, line 128: These files should be added in a single changeset, so you'll have to update the `file_added_to_ancillary_trust` event and associated handler to handle multiple file paths as an argument.  Similar to my comment from the add/delete trust in the ATDA file.

dorschs57 (changes requested):

dorschs57:
Here are just some general comments on this:

1. The code in the system and ancillary trust database admin pages should be smart enough that I’m not creating a changeset for a file that doesn’t actually change its trust status. For example right now I can multiselect a mixture of trusted and untrusted files. My options on which button (Trust or Untrust) that is active is dictated by the last selected file. If the last file selected was untrusted, I can untrust the whole set of selected file. This means I’m generating changesets to untrust files that are already untrusted. My suggestion would be to only allow an action to take place if all the selected files have the same status. i.e. if they are all untrusted the Trust button is active, if they are all trusted the Untrust button is active, and if they are mixed both button are disabled. Apply similar logic to the system trust page and the Add to Ancillary Trust button.

2. Multiselect on the Policy Analysis page is useless at this point. We can only display data for a single selected item, and the only action, Reconciliation, doesn’t allow for multi select. Here my suggestions. Eliminate the multiselect on the users and groups. There is no actions that can be taken on them right now and merging multiple users/groups subject list together seems complicated. I don’t know that we want to do that. We can keep the multiselect on subjects and objects lists for now with the idea of the reconciliation wizard to come down the road. I created issue #364 for this.

dorschs57:
> 2. Multiselect on the Policy Analysis page is useless at this point. We can only display data for a single selected item, and the only action, Reconciliation, doesn’t allow for multi select. Here my suggestions. Eliminate the multiselect on the users and groups. There is no actions that can be taken on them right now and merging multiple users/groups subject list together seems complicated. I don’t know that we want to do that. We can keep the multiselect on subjects and objects lists for now with the idea of the reconciliation wizard to come down the road. I created issue Policy Analysis Reconciliation Wizard #364 for this.

After talking with @jw3 I'll amend this a bit.  I'd still get rid of multiselect on users and groups.  On the subjects and objects for this PR:
1. If the user single selects and right-clicks, keep as is with the Reconcile option
2. If the user multi-selects and right-clicks, remove the Reconcile option from the context menu and replace it with an option to Trust or Untrust (depending on the trust status of the selected files).  If the user selects Trust/Untrust, show a simple `Yes/No` dialog to confirm and then generate a changeset for all selected items. This would be more of a "bulk action" compared to a wizard.  Once #364 is in place we can swap it out for that.

jw3:
Change the text of the button on the System Trust tab to be "Add To Ancillary Trust Database", rather than picking singular or plural version of "Files"

dorschs57 (changes requested):

fapolicy_analyzer/ui/system_trust_database_admin.py
```diff
@@ -86,41 +86,54 @@ def on_next_system(self, system):
...
+                addBtn.set_sensitive(False)
+                return
+
             addBtn.set_sensitive(not trusted)
 
-            self.trustFileDetails.set_in_database_view(
-                f(
-                    _(
-                        """File: {trust.path}
+            for trust in trusts:
```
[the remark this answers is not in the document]
egbicker: You're right. I thought this was doing more than just displaying which is why I left it in. Same for the ancillary one.

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -164,12 +176,41 @@ def on_view_row_activated(self, treeView, row, *args):
...
+                self.fileChangeContextMenu.remove(self.fileChangeContextMenu.trustItem)
+                self.fileChangeContextMenu.remove(self.fileChangeContextMenu.untrustItem)
+                if matching:
+                    if trust[0] == "u":
+                        self.fileChangeContextMenu.append(self.fileChangeContextMenu.trustItem)
+                    else:
+                        self.fileChangeContextMenu.append(self.fileChangeContextMenu.untrustItem)
+
+                self.fileChangeContextMenu.show_all()
+                self.fileChangeContextMenu.popup_at_pointer()
```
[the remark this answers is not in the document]
egbicker: I like your fix. The other option would be to have a "Trust status mismatch" menu option pop up instead and have it do nothing but provide info.

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -124,23 +125,23 @@ def __handle_selection_changed(self, data):
...
-            file = subject.file
-            trustList = (
-                self._systemTrust
-                if trust == "st"
-                else self._ancillaryTrust
-                if trust == "at"
-                else []
-            )
-            return first_true(trustList, pred=lambda x: x.path == file)
+    def find_db_trust(self, subject):
```
dorschs57, line 128: Make this a "private" method by using the double underscore prefix.

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -161,15 +162,14 @@ def __show_confirmation_dialog(self, subjects):
...
         if confirm == 1:
             changeset = Changeset()
             paths = [subject.file for subject in subjects]
-            if trust == "u":
-                changeset.add_trust(*paths)
-            else:
+            if isinstance(trust, Trust):
                 changeset.del_trust(*paths)
+            else:
+                changeset.add_trust(*paths)
```
dorschs57, line 172: I still don't think you logic is quite right here.  Not that you have the trust you need to look at the `trust.status` to determine the action.
* `trust` is None = Add
* `trust` is `t` and `subject.trust` is `st` = no action (there is nothing we can change about a file trusted in the system db)
* `trust` is `t` and `subject.trust` is `at` = Del
* `trust` is not `t` = Add

You can look at `TrustReconciliationDialog.__load_details` for how I'm setting the button visibility for this.

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -141,6 +155,25 @@ def find_db_trust(subject):
...
+        confirmationDialog.destroy()
+
+        trust = self.find_db_trust(subjects[0])
+        if confirm == 1:
+            changeset = Changeset()
+            paths = [subject.file for subject in subjects]
+            if isinstance(trust, Trust):
+                changeset.del_trust(*paths)
+            else:
+                changeset.add_trust(*paths)
```
dorschs57, line 172: I still don't think you logic is quite right here. Not that you have the trust you need to look at the trust.status to determine the action.

trust is None = Add
trust is t and subject.trust is st = no action (there is nothing we can change about a file trusted in the system db)
trust is t and subject.trust is at = Del
trust is not t = Add
You can look at TrustReconciliationDialog.__load_details for how I'm setting the button visibility for this.

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -141,6 +155,25 @@ def find_db_trust(subject):
...
             dispatch(apply_changesets(changeset))
 
+    def __show_confirmation_dialog(self, subjects):
+        changeset = None
+        parent = self.get_ref().get_toplevel()
+        confirmationDialog = ConfirmChangeDialog(parent=parent).get_ref()
+        confirm = confirmationDialog.run()
+        confirmationDialog.destroy()
+
+        trust = self.find_db_trust(subjects[0])
```
dorschs57, line 165: I think here you want to loop over all the subjects and add them to the `Changeset` right?

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -164,12 +197,41 @@ def on_view_row_activated(self, treeView, row, *args):
...
+                )
+                if len(trust) > 1:
+                    return
+
+                self.fileChangeContextMenu.remove(self.fileChangeContextMenu.trustItem)
+                self.fileChangeContextMenu.remove(self.fileChangeContextMenu.untrustItem)
+                if isinstance(trust[0], Trust):
+                    self.fileChangeContextMenu.append(self.fileChangeContextMenu.untrustItem)
+                else:
+                    self.fileChangeContextMenu.append(self.fileChangeContextMenu.trustItem)
```
dorschs57, line 221: Similar to up in line 169, you need to use both the `subject.trust` and the `dbtrust.status` values here to determine what menu option to show.

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -164,12 +197,41 @@ def on_view_row_activated(self, treeView, row, *args):
...
-            self.contextMenu.popup_at_pointer()
+            if n_paths == 1:
+                self.reconcileContextMenu.popup_at_pointer()
+            elif n_paths > 1:
+                trust = list(
+                    {
+                        self.find_db_trust(model.get_value(model.get_iter(path), 3))
+                        for path in iter(pathlist)
+                    }
+                )
```
dorschs57, line 212: This logic is harder than I was first thinking, and this isn't going to work.  You need to but some logic in to determine the  list of actions (trust or untrust) based on both the `subject.trust` (AT, ST, U) and the `dbtrust.status` (T, D).  Then check if all the actions are the same.  If not, exit out without giving the user to any menu options.

dorschs57 (changes requested):

dorschs57:
I submitted a couple easy fixes from my testing. 
1. I lowered the min GTK version on the confirm dialog to 3.22.  This is the version we use everywhere.
2. I returned `True` from the right-click handler of the `subjectList`.  This stops the event cascading so when you have multiple files selected and right-click it keeps all the files selected.

dorschs57 (changes requested):
I'm seeing some issues with files trusted in the system trust database. I'm testing with the `fapolicy-analyzer/tests/data/events0.log` file. 
1. Select `bin` user
2. Multiselect `/usr/bin/find` and `/usr/bin/gzip`
3. Right-click and I'm presented with the `Untrust Files` option even though both are trusted in the system trust db. (issue 1)
4. Select `Untrust Files` and the confirm dialog says `2 files will be trusted. Trust selected files?` even though the option was to Untrust the files. (issue 2)
5. Click `Yes` to confirm and it appears to create an empty changeset. (issue3)

PR #363, opened 2021-12-22
Provide TDU in analysis results

### LINKED ISSUE #328: Incorporate trust status with analysis results
Analysis results currently return the trust source (ST/AT/U) which is not enough information to fully render the status in the UI.  The analysis should also return the trust status (T/D/U) which will indicate whether the trust is valid for the current state of the system.

jw3:
This change adds the TDU status into the analysis results.

Closes #328

### DIFF
--- a/crates/analyzer/src/events/analysis.rs
+++ b/crates/analyzer/src/events/analysis.rs
@@ -13,6 +13,7 @@ use crate::error::Error::AnalyzerError;
 use crate::events::db::DB as EventDB;
 use crate::events::event::{Event, Perspective};
 use fapolicy_rules::Decision::*;
+use fapolicy_trust::stat::Status::{Discrepancy, Missing, Trusted};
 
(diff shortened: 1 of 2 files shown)


jw3:
Yet another one that needs #181

:gift:  :christmas_tree: :gift: 
:santa: :thinking:

dorschs57:
This will allow the Policy Analysis page to get closer to what we need with.  We can now color one of the sources of trust correctly, but we can't handle trust from multiple sources.  Is that what #181 is about?  For example a file that is untrusted in the ST but trusted in the AT should be bold/red ST and bold/green AT.

jw3:
>  For example a file that is untrusted in the ST but trusted in the AT should be bold/red ST and bold/green AT.

This is tricky.  The trust store contains a single entry, if it is AT there is no way to know if it is an overwrite or not.  We could make a guess based on other things like rpm db, but this is leaky.

PR #367, opened 2021-12-28
Event timestamp parsing

### LINKED ISSUE #289: Analysis should have access to event time information
It is important to be able to read event logs generated by `journalctl` but we should keep time information in these logs so we can show _the most recent state_ in our analysis.

If we're running on the same system that we want to analyze, then supporting a feature to read _directly from the tap_ (e.g, popen(3) journalctl) would be great as we don't need to first dump logs to a file to read them in.  `journalctl` also takes time ranges so we can pull events for analysis within time ranges more dynamically.

jw3:
Optionally parse event time from syslog events.

Keeping it optional to support debug logs which have no timestamp.

Closes #289


dorschs57 (approved):
Should there be a corresponding UI ticket to do something with this timestamp if its available?

jw3:
@dorschs57  yeah, though I dont have a definition of "do something" right now.

This PR sets us up for a temporal filter or some other good first use case.

Whatever it is, be sure to keep in mind that the timestamp is optional.

PR #375, opened 2022-01-13
Fix rust clippy lint warning

tparchambault:
1/13/2022 rust tool update generates a CI test suite failure (clippy lint warning).


tparchambault:
Tested on an FC34 VM via rpm installation. No issues observed.

jw3:
https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1580-2022-01-13

PR #377, opened 2022-01-17
Improve app version generation

### LINKED ISSUE #178: Populate the About screen with build info
Build info in the about screen would be useful for devs and end users.

- Build ID (probably git describe)
- ~~Current changelog~~

jw3:
Adds a pep440 compliant version capture to the build. Sets this in the package build and exposes it to the application by setting the module version number on the fapolicy_analyzer module.

Sets the About dialog version field to the module version.

Uses a modified CC0 installation of [versioneer](https://github.com/python-versioneer/python-versioneer) to provide the version capture.

Note that this is a separate concern from the RPM version calculation, which has a different schema. 

Closes #178


tparchambault (approved):

tparchambault:
On a clean FC34 system via rpm install, using the rpm associated with the CI RPM generation Action, I observed the version being 0.0.0, and no buttons on the About.... dialog to dismiss it. Can we get something into the version string for the PR rpms? timestamp?

jw3:
Yeah it wasn't quite ready, all wired now.

image

No idea why you don't see close.

jw3:
For tags

image

Also for dirty dev environment

image

dorschs57 (approved):

tparchambault:
Info box versioning information appears correct on a clean FC34 system installing the rpm from the latest CI run. No acknowledgement button widgets however; the info dialog can be closed via Escape nonetheless. This is a known issue independent of the work associated with this PR.

PR #378, opened 2022-01-19
Add make based development command driver

tparchambault:
First iteration to capture the standard developer workflow commands.


Makefile
```diff
@@ -0,0 +1,86 @@
...
+	pipenv run cargo fmt
+
+lint: pylint rslint
+	@echo "--- Source linting completed"
+
+pylint:
+	@echo "  |--- Python linting..."
+	pipenv run flake8 fapolicy_analyzer/
+
+rslint:
```
jw3, line 78: Suggest naming this `clippy`

Makefile
```diff
@@ -0,0 +1,86 @@
...
+	@echo "  |--- Rust linting..."
+	pipenv run cargo clippy --all
+
+
+pushprep: format lint test
```
jw3, line 83: `check` seems like a better name here to me

Makefile
```diff
@@ -0,0 +1,86 @@
...
+# fapolicy-analyzer execution
+run: dev-env
+	pipenv run python -m fapolicy_analyzer.ui
+
+run-verbose: dev-env
+	pipenv run python -m fapolicy_analyzer.ui -v
+
+###############################################################################
+# Development unit-testing and source code style tools
+test:
```
jw3, line 58: Suggest breaking the `test` target down into rust and python pieces like the lint target.

Makefile
```diff
@@ -0,0 +1,86 @@
...
+pybindings: pipenv-install
+	@echo "  |--- Generating python bindings..."
+	pipenv run python setup.py develop
+
+###############################################################################
+# fapolicy-analyzer execution
+run: dev-env
+	pipenv run python -m fapolicy_analyzer.ui
+
+run-verbose: dev-env
```
jw3, line 53: Don't use a separate target here, just allow an env var to be set.

A common `make` pattern found in the wild looks like

`V=1 make build`

Makefile
```diff
@@ -0,0 +1,86 @@
...
+	@echo
+	@echo "     list    - Display common development targets"
+	@echo "     dev-shell  - Install deps, build bindings, start venv shell"
+	@echo "     run     - Execute the fapolicy-analyzer"
+	@echo "     listall - Display all targets"
+	@echo "    "
+
+###############################################################################
+# Development environment population and creation
+dev-shell: pipenv-install pybindings
```
jw3, line 32: Some general renaming suggestions

`dev-shell` -> `shell`
`dev-env` -> `env`
`pipenv-install` -> `install`
`pybindings` -> `build`

Makefile
tparchambault: @jw3 I'll make it whatever you'd like, however my standard pushback would be does the standard developer on the street know what clippy is? They might know what lint means and then can infer that clippy would be a rust linting tool. Ditto cargo; ditto black; ditto flake.

Makefile
tparchambault: I like that.

Makefile
tparchambault: Do we have rust unit-testing? afaik we are only running the pytest suite.

Makefile
tparchambault: This was first pass. Things are still being worked. I'd prefer to pass and handle a CL arg. but I get your point.

Makefile
tparchambault: Ok.

Makefile
jw3: I'd guess only a Rust developer would be interested in linting Rust, in which case they would know what `clippy` is.  The `rslint` target however would be an unknown to everyone.

Makefile
jw3: Sure, the `Test Suite` workflow steps is all Rust all the time.

`cargo test --all`

Makefile
```diff
@@ -0,0 +1,86 @@
...
+	@echo "--- Invoking pytest unit-test suite"
+	pipenv run xvfb-run -a pytest -s --cov fapolicy_analyzer/
+
+format: pyformat rsformat
+	@echo "--- Invoking formatting tools"
+
+pyformat:
+	pipenv run black fapolicy_analyzer/
+
+rsformat:
```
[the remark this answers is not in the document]
tparchambault: Done

Makefile
tparchambault: Ya but any developer on the team is subject to those checks in the CI pipeline.. Always easier to find these failures locally. For instance, I've discovered and fixed two issues related to rust tool updates but only after pushing to github and having the CI pipeline fail, and really what do I know about rust. Effectively, zilch.

tparchambault: Anyhoo, I'm going with your naming suggestions.

Makefile
tparchambault: Done. Will push shortly.

Makefile
tparchambault: Yea, I saw that in main.yml. Thank you. I'll add it to the Pipfile aliases, so that we can exercise them pre-push.

Makefile
tparchambault: Done. Will push shortly.

jw3: Members of the team aren't "standard developers on the street" though.  I would expect the team to know what clippy is.

Makefile
tparchambault: Done, will push shortly.

Makefile
tparchambault: Done. Push shortly.

Makefile
tparchambault: Added the ability to pass in CL arguments via the OPTIONS environment variable when using the 'run' target, e.g.
`$ OPTIONS='-v -h' make run`

tparchambault:
Added output messaging. Ideally would like to redirect stdout so that the output appears like the boot init messages with command failures displayed at the point of failure. Full output would be available via an option to increase verbosity. Next go round, when things are slow...

tparchambault:
@jw3 , the python file changes were the result of executing the 'black' formatter.

jw3 (approved):

PR #380, opened 2022-01-21
Revise user and group lookup

### LINKED ISSUE #349: Revise lookup of users and groups
- do not read directly from `passwd` or `group`
- support LDAP

jw3:
Using `getent` rather than reading directly from etc db files ensures that we pull all users and groups from the Name Service Switch libraries rather than from the local machine config only.

Closes #349


jw3:
@tparchambault can you give this a test, I was seeing an issue, need a second set of eyes, or at least half a second set :wink:

tparchambault (approved):
I was surprised that the absolute path to getent wasn't needed by std::process::Command() but I did see that the PATH is inherited, so looks good to go.

Performing cursory testing over FC34.

Issue #234, opened 2021-09-15
Add rust fmt config


jw3:
https://rust-lang.github.io/rustfmt

tparchambault:
Did you want anything specific in the toml file? 

Fwiw, I had no luck, enabling 'unstable-features' using either with the `unstable_features = true` config file directive, or the CL option `--unstable-features` passed to `rustfmt` as stated in the link above.

However, if you do want to leverage the unstable features, what appears to work is to add the `+nightly` option to `cargo fmt` while using the config file directive `unstable_features = true`

``
$ cargo +nightly fmt
``

https://github.com/rust-lang/rustfmt/issues/2227 was useful wrt the +nightly option approach. There may be something more current in the wild.

jw3:
Closing as we have no fmt preferences at this time.

PR #381, opened 2022-01-24
Add rustfmt.toml config file

tparchambault:
Empty rustfmt.toml config file added. Tested w/unstable features and verified source was reverted when those same features were removed from the new toml file.


tparchambault:
We are not supporting unstable-features; Removed the `+nightly` option from the cargo fmt command in the Makefile.

jw3:
I think for now we can close this and #234 until there are changes to the default fmt config.

PR #395, opened 2022-02-03
Refresh view on changeset application

egbicker:
Checks the current number of changesets and compares it to the last known number. If there is a difference, refresh the policy analysis view.

### DIFF
--- a/fapolicy_analyzer/ui/policy_rules_admin_page.py
+++ b/fapolicy_analyzer/ui/policy_rules_admin_page.py
@@ -68,6 +68,7 @@ class PolicyRulesAdminPage(UIConnectedWidget, UIPage):
         }
         UIPage.__init__(self, actions)
 
+        self.__n_changesets = 0
         self.__audit_file: Optional[str] = audit_file
         self.__log: Optional[Sequence[EventLog]] = None
         self.__events_loading = False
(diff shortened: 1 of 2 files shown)


fapolicy_analyzer/ui/policy_rules_admin_page.py
```diff
@@ -437,6 +438,11 @@ def exec_primary_data_func():
             self.__events_loading or self.__users_loading or self.__groups_loading
         )
 
+        if system.get("changesets"):
+            if not self.__n_changesets == len(system.get("changesets")):
+                self.__n_changesets = len(system.get("changesets"))
+                self.__refresh()
```
dorschs57, line 444: to avoid the nested if's here I'd suggest something like 

```suggestion
        num_changesets = len(system.get("changesets",[]))
        if self.__n_changesets != num_changesets:
            self.__n_changesets = num_changesets
            self.__refresh()
```

dorschs57: If you'd like the try the walrus out something like this works too 😃 

```suggestion
        if (num_changesets := len(system.get("changesets",[]))) != self.__n_changesets:
            self.__n_changesets = num_changesets
            self.__refresh()
```

dorschs57 (approved):
I tested out the functionality and its working as expected.  I had the one comment on the code if you wish to change it.

PR #403, opened 2022-02-16
Fixup: Fix rpm build issue

dorschs57:
Updated the URL used for pulling the GPG key for the Appstream repo during RPM builds on RHEL 8.5. The old URL was breaking the builds with a bad key https://github.com/ctc-oss/fapolicy-analyzer/runs/5218341144?check_suite_focus=true.  The new URL is pulling the key from the same location we are pointing to for the Rocky 8.5 Appstream repo `baseurl`.

### DIFF
--- a/.github/workflows/rpm.yml
+++ b/.github/workflows/rpm.yml
@@ -1,7 +1,7 @@
 on:
   push:
     branches:
-      - master
+      - fix-rpm-build-issue
     tags:
       - 'v*'
   pull_request:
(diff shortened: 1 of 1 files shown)


jw3:
How about updating the description of the PR with something describing this and a link to reference, if there is one.

tparchambault:
Same issue with the url embedded in the .../fapolicy-analyzer/scripts/rpm/Dockerfile. Used @dorschs57 fix's pub version of the url addressed my issue w/downloading from the RockyLinux repo.

PR #406, opened 2022-02-23
Updated Rules Administration Text Editor to read directly from the fapolicyd rules file

dorschs57:
This PR reworks the Rules Admin tool's Text Editor view so it reads in and displays the fapolicyd rules file directly instead of using the parsed rules list from the backend. Right now it's hard coded to the monolithic rules file location. The backend will need updates to handle both the rules.d directory structure and a call to return the current rules file(s) location as config data at some point.

### DIFF
--- a/fapolicy_analyzer/ui/rules/rules_text_view.py
+++ b/fapolicy_analyzer/ui/rules/rules_text_view.py
@@ -13,9 +13,11 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-from typing import Sequence
+import logging
 
-from fapolicy_analyzer import Rule
+from fapolicy_analyzer.ui.actions import NotificationType, add_notification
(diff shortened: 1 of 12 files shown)


fapolicy_analyzer/ui/rules/rules_text_view.py
```diff
@@ -25,6 +27,14 @@ def __init__(self):
         self.__text_view = self.get_object("textView")
         self.__text_view.set_show_line_numbers(True)
 
-    def render_rules(self, rules: Sequence[Rule]):
-        text = "\n".join([r.text for r in rules])
+    def render_rules(self, rules_path: str):
+        text = ""
+        print(open)
```
egbicker, line 32: Does this still need to be here?

dorschs57: Good catch. Left over from getting the tests running. It gone now.

egbicker (approved):

PR #408, opened 2022-03-01
Policy Analysis Trust column text color updates

### LINKED ISSUE #389: Color code the Subject and Object trust status in the Policy Event Analysis tool
The `ST/AT/U` in the Trust column of the Subject and Object lists should be color coded based on the current trust status of the file. I believe this is now possible following #363 since the trust status is now returned as part of the the event object.  Based on the prototype document the rules for coloring should be:

1. trusted in ST = ST is bold, green, and underlined
2. untrusted in ST = ST is bold and red
3. trusted in AT = AT is bold, green, and underlined
4. untrusted in AT = AT is bold and red
5. unknown trust = U is bold, green, and underlined (does this make more sense to be bold and red?)

These can be combined together as well.  For example if the file is untrusted in ST but trusted in AT the trust line would be bold/red ST and bold/green/underlined AT.  Reference page 3 of the prototype document for more details.

egbicker:
closes #389

### DIFF
--- a/fapolicy_analyzer/tests/test_object_list.py
+++ b/fapolicy_analyzer/tests/test_object_list.py
@@ -70,11 +70,11 @@ def test_loads_store(widget):
 
 def test_status_markup(widget):
     view = widget.get_object("treeView")
-    st_red = '<span color="red"><b>ST</b></span>'
-    at_red = '<span color="red"><b>AT</b></span>'
-    u_green = '<span color="green"><u><b>U</b></u></span>'
-    st_green = '<span color="green"><u><b>ST</b></u></span>'
-    at_green = '<span color="green"><u><b>AT</b></u></span>'
+    st_red = f'<span color="{Colors.RED}"><b>ST</b></span>'
+    at_red = f'<span color="{Colors.RED}"><b>AT</b></span>'
+    u_green = f'<span color="{Colors.GREEN}"><u><b>U</b></u></span>'
+    st_green = f'<span color="{Colors.GREEN}"><u><b>ST</b></u></span>'
+    at_green = f'<span color="{Colors.GREEN}"><u><b>AT</b></u></span>'
 
     # System trust
     widget.load_store([_mock_object(trust="ST", trust_status="U")])
--- a/fapolicy_analyzer/tests/test_object_list.py
+++ b/fapolicy_analyzer/tests/test_object_list.py
@@ -71,24 +71,30 @@ def test_loads_store(widget):
 def test_status_markup(widget):
     view = widget.get_object("treeView")
 
+    st_trust = '<span color=\"green\"><b><u>ST</u></b></span>'
+    at_trust = '<span color=\"green\"><b><u>AT</u></b></span>'
+    st_untrust = '<span color=\"red\"><b>ST</b></span>'
+    at_untrust = '<span color=\"red\"><b>AT</b></span>'
+    untrust = '<span color="green"><b><u>U</u></b></span>'
+
     # System trust
     widget.load_store([_mock_object(trust="ST")])
-    assert view.get_model()[0][0] == "<b><u>ST</u></b>/AT/U"
+    assert view.get_model()[0][0] == st_trust + "/" + at_untrust + "/U"
     # Ancillary trust
     widget.load_store([_mock_object(trust="AT")])
-    assert view.get_model()[0][0] == "ST/<b><u>AT</u></b>/U"
+    assert view.get_model()[0][0] == st_untrust + "/" + at_trust + "/U"
     # Untrusted
     widget.load_store([_mock_object(trust="U")])
-    assert view.get_model()[0][0] == "ST/AT/<b><u>U</u></b>"
+    assert view.get_model()[0][0] == st_untrust + "/" + at_untrust + "/" + untrust
     # Bad data
     widget.load_store([_mock_object(trust="foo")])
-    assert view.get_model()[0][0] == "ST/AT/U"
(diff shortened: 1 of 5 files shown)


fapolicy_analyzer/tests/test_object_list.py
```diff
@@ -71,24 +71,30 @@ def strip_markup(markup):
 def test_status_markup(widget):
     view = widget.get_object("treeView")
 
+    st_trust = '<span color=\"green\"><b><u>ST</u></b></span>'
+    at_trust = '<span color=\"green\"><b><u>AT</u></b></span>'
+    st_untrust = '<span color=\"red\"><b>ST</b></span>'
+    at_untrust = '<span color=\"red\"><b>AT</b></span>'
+    untrust = '<span color="green"><b><u>U</u></b></span>'
```
dorschs57, line 78: If you are using single quotes on the outside, you don't need to escape the double quotes on the inside. There are a few places where this needs changed.

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -106,7 +105,22 @@ def _columns(self):
...
+        u_str = "U"
+        at_str = '<span color=\"red\"><b>AT</b></span>'
+        st_str = '<span color=\"red\"><b>ST</b></span>'
+
+        if subject.trust.lower() == "at":
+            at_str = '<span color=\"green\"><b><u>AT</u></b></span>'
+        elif subject.trust.lower() == "st":
+            st_str = '<span color=\"green\"><b><u>ST</u></b></span>'
+        elif subject.trust.lower() == "u":
+            u_str = '<span color=\"green\"><b><u>U</u></b></span>'
```
dorschs57, line 118: Its not enough here to just look at the `subject.trust`. You need to also look at the `subject.trust_status`. The `trust` only tells which trust database the file appears in. The `trust_status` tells if it is trusted in that database or not.

dorschs57 (changes requested):

jw3:
@dorschs57 could you start a design discussion with what you just described and a snip from the prototype doc?

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -107,23 +107,36 @@ def _columns(self):
...
-        at_str = '<span color=\"red\"><b>AT</b></span>'
-        st_str = '<span color=\"red\"><b>ST</b></span>'
+        at_str = "AT"
+        st_str = "ST"
+
+        status = (
+            dbtrust.status.lower()
+            if (dbtrust := self.__find_db_trust(subject))
+            else None
+        )
```
dorschs57, line 117: You shouldn't need to do this any more to get the status. #363 added the `trust_status` property to the `subject` and `object`. At some point we should go back and update the older code that is doing it this way too...when time permits.

dorschs57 (changes requested):

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -106,29 +106,21 @@ def _columns(self):
...
-            if not status or status == "d":
-                st_str = '<span color="red"><b>ST</b></span>'
-            else:
-                st_str = '<span color="green"><u><b>ST</b></u></span>'
-        elif subject.trust.lower() == "u":
-            if not status or status == "d":
-                u_str = '<span color="green"><u><b>U</b></u></span>'
+        status = subject.trust_status.lower()
+
+        if status == "u":
```
dorschs57, line 112: I think this should be `status == 'd'` right?

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -106,29 +106,21 @@ def _columns(self):
...
+            at_str = '<span color="red"><b>AT</b></span>' if subject.trust.lower() == "at" else "AT"
```
dorschs57, line 113: I'd suggest you assign the `subject.trust.lower()` to variable like you did for the status above and reuse it.

fapolicy_analyzer/ui/subject_list.py
```diff
@@ -106,29 +106,21 @@ def _columns(self):
...
+            st_str = '<span color="red"><b>ST</b></span>' if subject.trust.lower() == "st" else "ST"
```
dorschs57, line 114: We have a `Colors` class in the `configs.py` file. Add the "red" and "green" colors to that and reference the constants instead of the hard coded string.

dorschs57 (changes requested):

fapolicy_analyzer/ui/subject_list.py
egbicker: When I switched to using trust status it was always a "u" or "t". Was something changed when `trust_status` was added?

dorschs57: According to this guy https://github.com/ctc-oss/fapolicy-analyzer/blob/ea154174a5b9b94ce1253af45756e71151d3d2b5/crates/analyzer/src/events/analysis.rs#L94 the `u` is for unknown file trust and `d` is for discrepancy. I'm not exactly sure how a file could be in one of the trust databases and still have a `u` for status, but maybe the best approach would would be to color it red if it's status is `not 't'`.

egbicker: Good call. Fixed.

jw3:
> I'm not exactly sure how a file could be in one of the trust databases and still have a u for status

If the file does not exist on disk.

dorschs57:
> > I'm not exactly sure how a file could be in one of the trust databases and still have a u for status
> 
> If the file does not exist on disk.

I'm thinking it should still be colored red right?

dorschs57 (approved):

PR #433, opened 2022-03-21
Application profiling integration

tparchambault:
Do not review / merge. Need rpm generation to work as su to programatically control fapd.


tparchambault:
Added the env var, FAPD_LOGPATH to override default profiler log file names w/embedded timestamps. Log file names will get ".stdout" and ".stderr" suffixes appropriately appended to $FAPD_LOGPATH,

tparchambault:
Keep the branch pls.

PR #435, opened 2022-03-23
Display rule status and validation in the guided editor

### LINKED ISSUE #401: Provide status information in the rules editor
Provide a status information area at the bottom of the rules editor. This would be displayed for both the list view and the monolithic view of the rules. What information is provided by the status area will be dependent on what information we can get from the backend.

### LINKED ISSUE #432: Gtk Warning about colors on policy analysis view
Open an event log in policy analysis view and navigate to the object view.  If there are denies present you get a warning on the console

```
(__main__.py:13341): Gtk-WARNING **: 00:52:01.505: Failed to set text from markup due to error parsing markup: Value of 'foreground' attribute on <span> tag on line 1 could not be parsed; should be a color specification, not 'FF0000'
```

dorschs57:
Adds collapsible rows to the Guided Editor view that can be expanded to who any rule metadata. The rules are also colored according to the metadata.

Closes #401


jw3 (approved):

jw3:
@dorschs57 The collapsible rows and coloring look great.

PR #436, opened 2022-03-23
Rules.d support

### LINKED ISSUE #428: Load rules.d
Load `rules.d` files into a single rules structure that contains metadata to support source context aware operations. 

It is worth noting that the fapolicyd project already has a script (**[fagenrules]**) for this that takes the files in rules.d and builds a single file (`compiled.rules`) that the daemon ultimately ingests ingests.  For us, reading the rules post-compilation does not support the things like the ability to edit the individual files or give feedback with source file context on validation errors.  So we need to pull these files in on our own and preserve and generate the appropriate metadata.

The compiled rules can be used as a data input to integration testing to ensure we are interpreting the d structure in the same way as fapolicyd would expect.

[fagenrules]: https://github.com/linux-application-whitelisting/fapolicyd/blob/main/init/fagenrules

jw3:
Add support for specifying a directory as the source of rules.

When specified the loader reads all *.rules files under this dir into the Rules DB.

Closes #428

### DIFF
--- a/crates/pyo3/src/system.rs
+++ b/crates/pyo3/src/system.rs
@@ -14,7 +14,6 @@ use fapolicy_analyzer::events::db::DB as EventDB;
 use fapolicy_app::app::State;
 use fapolicy_app::cfg;
 use fapolicy_app::sys::deploy_app_state;
-use fapolicy_daemon::fapolicyd;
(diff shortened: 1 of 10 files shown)


jw3:
image

PR #440, opened 2022-03-24
Initial rule linting

jw3:
Rule linting reference implementation intended to providing substance for  testing, discussion, and seeing the full round trip to the frontend.  A better solution will emerge.

### lints

image

- warn: when an `any+all+all` rule is not the last rule
- warn: when a subject exe does not exist on disk
- warn: when an object path does not exist on disk


jw3:
The failing UBI 8  build pointed out the need to set MSRV
- #441

PR #445, opened 2022-03-25
Set Minimum Supported Rust Version

### LINKED ISSUE #441: Set minimum supported Rust version (MSRV)
The first linting implementation raised an issue with a language feature that is available in all environments except RHEL 8.

- [`error[E0658]: pattern bindings after an `@` are unstable`](https://github.com/ctc-oss/fapolicy-analyzer/runs/5671149650?check_suite_focus=true#step:10:492)

The binary would run fine on RHEL 8, but we are also required to build packages there.

### Two options
1. Allow unstable Rust
2. Pin the MSRV to unify all environments

Since this is the only feature, and the functionality can be redesigned, we will do option 2 and set the MSRV to ensure builds can run on RHEL 8.

We are limited to **1.54**
- [`rust                        x86_64  1.54.0-3.module+el8.5.0+13074+d655d86c   ubi-8-appstream       25 M`](https://github.com/ctc-oss/fapolicy-analyzer/runs/5671149650?check_suite_focus=true#step:9:41)

### Reference

- https://github.com/foresterre/cargo-msrv
- https://foresterre.github.io/cargo-msrv/commands/list.html#description

jw3:
Use package.metadata to pin the minimum Rust version we must support.

This is needed to ensure we can continue to build RPMs for RHEL where Rust 1.54.0 is available.

Closes #441


jw3:
There is no standard for maintaining msrv.  Took a minimal approach with #480 and will revisit in the future if standardized controls become available.

Issue #393, opened 2022-01-31
Right click should select


jw3:
With a single right the user should be able to select a list item and activate the context menu.

As it is now you have to left click to select and then right click to get context menu.

PR #457, opened 2022-04-04
Fa237 remove missing ancillary

egbicker:


tparchambault:
Test on a clean FC34 via rpm install, running as root: GUI and interaction are all good. How is the removed file in the GUI, removed from the ATDbase? Did you wire that up yet, and if so, what's the user's workflow to push the change?

tparchambault:
Getting following logging output and call stack:
```
==== AUTHENTICATING FOR org.freedesktop.policykit.exec ====
Authentication is needed to run `/usr/bin/env' as the super user
Authenticating as: Thomas Archambault (toma)
Password: 
==== AUTHENTICATION COMPLETE ====
DEBUG:root:Verbosity enabled.
DEBUG:root:Xdg absolute path: /root/.local/state/fapolicy-analyzer/FaCurrentSession.tmp
DEBUG:root:SessionManager::set_autosave_filename: /root/.local/state/fapolicy-analyzer/FaCurrentSession.tmp
fapolicy-analyzer v0.0.0+211.g01fbce6
DEBUG:redux._internal.feature:system
DEBUG:root:_PostInitCaller.__call__((), {})
DEBUG:root:system_feature::finish::system = <class 'app.System'>
DEBUG:root:distroying widget from class SplashScreen
DEBUG:root:disposing of subscription for class SplashScreen
DEBUG:root:_PostInitCaller.__call__((), {})
DEBUG:root:_PostInitCaller.__call__((), {})
DEBUG:root:_PostInitCaller.__call__((), {})
DEBUG:root:_PostInitCaller.__call__((), {})
DEBUG:root:SessionManager::detect_previous_session()
DEBUG:root:Search Pattern: /root/.local/state/fapolicy-analyzer/FaCurrentSession.tmp_*.json
DEBUG:root:Glob search results: []
DEBUG:root:on_update_daemon_status(ServiceStatus.FALSE)
DEBUG:root:start_daemon_monitor(): ServiceStatus.FALSE
DEBUG:root:Spawning monitor thread...
DEBUG:root:_monitor_daemon() executing
DEBUG:root:Thread=<Thread(Thread-1, started daemon 140533840750144)>, Running=True
DEBUG:root:__update_fapd_status(ServiceStatus.FALSE)
  File "/usr/lib/python3.9/site-packages/fapolicy_analyzer/ui/ancillary_trust_database_admin.py", line 83, in on_view_button_press_event
    iter_ = model.get_iter(next(iter(pathlist)))  # single select use first path
StopIteration
```

tparchambault:
Tested via rpm installation on an FC34 system. Issues associated with file removal initiated by the file's context menu in the ATDA tool view have been addressed.

fapolicy_analyzer/glade/remove_deleted_dialog.glade
```diff
@@ -0,0 +1,90 @@
...
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with this program.  If not, see <https://www.gnu.org/licenses/>.
+-->
+<interface>
+  <requires lib="gtk+" version="3.20"/>
```
dorschs57, line 20: All our other glade files require 3.22.

fapolicy_analyzer/ui/ancillary_trust_database_admin.py
```diff
@@ -130,11 +155,45 @@ def on_files_added(self, files):
...
+                removeDialog.destroy()
+                if resp == Gtk.ResponseType.APPLY:
+                    self.delete_trusted_files(*dne_list)
+                    treeView = self.trustFileList.get_object("treeView")
+                    treeView.get_selection().select_path(Gtk.TreePath.new_first())
+                    model, pathlist = treeView.get_selection().get_selected_rows()
+                    if model:
+                        child_model = model.get_model().get_model()
+                        iter_ = child_model.get_iter(next(iter(pathlist)))  # single select use first pat
+                        child_model.remove(iter_)
```
dorschs57, line 193: This code looks the same as what is in the `on_trustBtn_click` method. Can we abstract it to a helper method and DRY this up a bit?

dorschs57 (changes requested):

fapolicy_analyzer/ui/ancillary_trust_database_admin.py
egbicker: Good call. Got it condensed and cleaned up

fapolicy_analyzer/ui/remove_deleted_dialog.py
```diff
@@ -0,0 +1,34 @@
...
+            return ("s", "are") if count > 1 else ("", "is")
+
+        super().__init__()
+        if parent:
+            self.get_ref().set_transient_for(parent)
+
+        textView = self.get_object("removeInfo")
+        textBuffer = textView.get_buffer()
+
+        displayText = f(_("{len(deleted)} file{plural(len(deleted))[0]} cannot be found on disc"))
```
dorschs57, line 33: "cannot be found on the file system" would be more consistent with the working in the file details panes.

dorschs57 (commented):
When I right click

jw3:
> When I right-click a file when the application first opens, without first left clicking one, I get an error

Sounds related to #393

dorschs57:
On RHEL8 the dialog displays like 
image

dorschs57:
When I select a missing file and remove it, when the treeview refreshes all my other files are gone except the one I removed.

dorschs57:
Overall I find this new functionality a bit confusing. When I select a missing file my button option is Trust, which doesn't make sense since that is an "add" operation. We are also depending on the user knowing to right-click in order to remove it, when delete functionality is already something we have in place with the Untrust button.

I would suggest getting rid of the right click menu and dialog. I think if the user selects a file that has a `U` status we need to handle that and set the Untrust button active and the Trust button disabled.  For clarity we can change the text on the button from Untrust to Delete in that instance. Thoughts on this approach?  

Maybe a `U` status isn't enough of a check. Should it be `U` status and empty file path?

jw3:
@dorschs57 all good points.  The "Untrust" function means "remove" whether the file exists on disk or not, and the only way to get into a "U" status is if the file was trusted and now it does not exist on disk.

egbicker:
I think the "Trust" option showing up as an available option for a deleted file is a bug where the wrong selection is removed from the view so it stays as an option. I agree that trusting a deleted file is counterintuitive and doesn't make sense as a workflow.

What should the button display if multiple files are selected and there is a mixture of files that currently exist and used to exist?

jw3:
Wouldn't the "Untrust" text cover all cases?

egbicker:
Probably, I was thinking about Shawn's Untrust/Delete button text change.

jw3:
This is a little late to ask this but why is remove a special case?  Untrust seems to be the same whether an entry exists on disk or not.

dorschs57:
The bug is that we don't handle the `U` status correctly and I think the fallback is to show the Trust button.  I agree if we fix the bug the Untrust button would cover what we need I think.  I only mentioned changing the text from Untrust to Delete in the case of a `U` status because it might be more user friendly.  If we want to just fix the bug and keep the text as Untrust I'm good with that too.

jw3:
Ok, if it's just changing the text for clarity I'm ok with that. I do lean towards "remove" rather than "delete".

jw3:
If the selection is mixed I would leave as untrust.

Though you have to weigh if the value added by changing the text in the homogeneous case outweighs ambiguity of not changing it in the mixed case.

Also how much complexity is added to the code for this thing that may or may not make it clearer..

jw3:
I dont have strong opinion here btw, just sharing thoughts.

dorschs57:
When I select a missing file now, neither Trust nor Untrust is enabled for me.

egbicker:
Is the deleted file still trusted when both are greyed out? Trust should be greyed out if it doesn't exist on disc and if Untrust is greyed out then it should be Untrusted, I only changed the Trust button logic

dorschs57:
> Is the deleted file still trusted when both are greyed out? Trust should be greyed out if it doesn't exist on disc and if Untrust is greyed out then it should be Untrusted, I only changed the Trust button logic

The buttons are working for Trusted and Untrusted files, but for missing files this is what I see 
image

fapolicy_analyzer/ui/ancillary_trust_database_admin.py
```diff
@@ -130,11 +137,31 @@ def on_files_added(self, files):
...
         if self.selectedFiles:
-            self.add_trusted_files(*self.selectedFiles)
+            dne_list = [f for f in self.selectedFiles if not os.path.isfile(f)]
+            selectedFiles = [f for f in self.selectedFiles if f not in dne_list]
+            if len(dne_list) > 0:
+                removeDialog = RemoveDeletedDialog(deleted=dne_list).get_ref()
+                resp = removeDialog.run()
+                removeDialog.destroy()
+                if resp == Gtk.ResponseType.APPLY:
+                    self.delete_trusted_files(*dne_list)
```
dorschs57, line 147: This should never happen right? Being able to click the Trust button on a file that's missing.

fapolicy_analyzer/ui/remove_deleted_dialog.py
dorschs57: If you are keeping the current text at least change the spelling to "disk".

dorschs57 (changes requested):
The Untrust buttons now working properly for missing files.  The issue I see now that I can untrust the missing file even after its already been deleted. 
image

Once a missing file is marked as deleted, I shouldn't be able any action on it. The button should be grayed out.

Also you still need to run `pipenv run extract_messages` to update the .pot file.

egbicker:
From what I'm seeing 
File exists on disk: Trust and Untrust as normal
File does not exist on disk, is trusted in the ATDb: Untrust selectable
File does not exist on disk, is marked as deleted/untrusted: Neither option selectable

dorschs57 (approved):

PR #463, opened 2022-04-05
integration of FapdManager

tparchambault:
Not ready for review. Moving locks into FapdManager from MainWindow. Testing on clean FC34 via rpm install.


tparchambault:
Currently each fapd instance (profiling and online) can be independently controlled with separate menu items.. When the new FA Profiling tab is available the FapdManager class should enact mutually exclusive control of both fapd instances, such that zero or one and only one fapd instance may be executing at any given time.

tparchambault:
Pls do not merge until @jw3 can verify start/stop/monitoring of the online fapolicyd instance, i.e. things work as always. Technically, the API to control the Profiling instance has been implemented within this PR but this PR is about moving the existing interactions with fapolicyd (and Locks) into this manager class. Starting/Stopping the profiling instance is not wired to the UI layer currently.

fapolicy_analyzer/ui/fapd_manager.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def get_mode(self):
+        """This function will be removed from the public API, when integration
+        with the Profiling Tool tab is completed."""
+        logging.debug("FapdManager::get_mode()")
+        return self.mode
+
+    def get_profiling_timestamp(self):
+        logging.debug("FapdManager::get_profiling_timestamp()")
+        return self._fapd_profiling_timestamp
```
dorschs57, line 127: Consider using the `@properties` decorator for these getters/setters https://docs.python.org/3.6/library/functions.html#property

fapolicy_analyzer/ui/fapd_manager.py
```diff
@@ -0,0 +1,215 @@
...
+    def stop_profiling_fapd(self):
+        # ToDo: Add logic to that online and profiling instances are mutex
+        self.set_mode(FapdMode.PROFILING)
+        self._stop()
+
+    def status_profiling_fapd(self):
+        # ToDo: Add logic to that online and profiling instances are mutex
+        # ToDo: Consider returning a tuple with status and instance info
+        self.set_mode(FapdMode.PROFILING)
+        return self._status()
```
dorschs57, line 95: Instead of duplicating the same methods for both `ONLINE` and `PROFILING` should we define a single set that each take a `mode` parameter?  It could default to default to `ONLINE` too maybe?

fapolicy_analyzer/ui/fapd_manager.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def stop_profiling_fapd(self):
+        # ToDo: Add logic to that online and profiling instances are mutex
+        self.set_mode(FapdMode.PROFILING)
+        self._stop()
+
+    def status_profiling_fapd(self):
+        # ToDo: Add logic to that online and profiling instances are mutex
+        # ToDo: Consider returning a tuple with status and instance info
+        self.set_mode(FapdMode.PROFILING)
```
dorschs57, line 94: Should requesting the status change the `mode` of the manager?

dorschs57 (changes requested):

fapolicy_analyzer/ui/fapd_manager.py
```diff
@@ -0,0 +1,215 @@
...
+        # ToDo: Verify that FAPD_LOGPATH works in the pkexec environment
+        if os.environ.get("FAPD_LOGPATH"):
+            self.fapd_profiling_stdout = os.environ.get("FAPD_LOGPATH") + ".stdout"
+            self.fapd_profiling_stderr = os.environ.get("FAPD_LOGPATH") + ".stderr"
+        else:
+            self.fapd_profiling_stdout = None
+            self.fapd_profiling_stderr = None
+
+    # ############### Public Interface ###############
+    def start_online_fapd(self):
```
jw3, line 65: The `_fapd` suffix on this and the following methods seems redundant.

bin/fapolicy-analyzer
```diff
@@ -40,7 +40,10 @@ else
 fi
 
 # Execute the fapolicy-analyzer with heightened privileges
-pkexec  --disable-internal-agent env NO_AT_BRIDGE=1 DISPLAY=$DISPLAY XAUTHORITY=$XAUTHORITY python3 -m fapolicy_analyzer.ui "$@"
+pkexec  --disable-internal-agent env NO_AT_BRIDGE=1 \
+	DISPLAY=$DISPLAY\
```
jw3, line 44: Space before the backslash

jw3:
Tested the daemon control and did not encounter any issues.

fapolicy_analyzer/ui/fapd_manager.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+class FapdMode(Enum):
+    DISABLED = 0
+    ONLINE = 1
+    PROFILING = 2
+
+
+class FapdManager():
+    def __init__(self, bSuFapdControl=True):
```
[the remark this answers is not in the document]
tparchambault: Agreed...

fapolicy_analyzer/ui/fapd_manager.py
```diff
@@ -0,0 +1,215 @@
...
+        logging.debug("FapdManager::__init__(self)")
+        self._fapd_status = ServiceStatus.UNKNOWN
+        self._fapd_monitoring = False
+        self._fapd_ref = Handle("fapolicyd")
+        self._fapd_profiler_pid = None
+        self._fapd_control_enabled = bSuFapdControl
```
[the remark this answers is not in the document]
tparchambault: It will. Previously calls to the fapd backend functions were restricted at the MainWindow level by graying out the menu items for non-root users.

Wired it up so that self.mode can not be modified unless an env var, SU_OVERRIDE is set, so that we can modify the mode during development. Eventually SU_OVERRIDE should be removed when the project has a production release.

bin/fapolicy-analyzer
tparchambault: Fixed. Will push shortly...

fapolicy_analyzer/ui/fapd_manager.py
tparchambault: Removed `_fapd` suffix on public functions.

fapolicy_analyzer/ui/fapd_manager.py
tparchambault: TBD. The original intention of the mode argument was to communicate with either fapolicyd instance with the same interface, so technically it would be possible for the client programmer to sequentially query the two fapolicyd instances. e.g. fapd status, and update the GUI appropriately. But... I'm not sure if that's the way to go. Will code up the control dlg and check the workflow.

bin/fapolicy-analyzer
tparchambault: Done

fapolicy_analyzer/ui/fapd_manager.py
tparchambault: Done; Pushed.

fapolicy_analyzer/ui/fapd_manager.py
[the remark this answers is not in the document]
tparchambault: Done; Pushed.

fapolicy_analyzer/ui/fapd_manager.py
```diff
@@ -54,41 +58,43 @@ def __init__(self, bSuFapdControl=True):
         self.initial_daemon_status()
 
         # ToDo: Verify that FAPD_LOGPATH works in the pkexec environment
-        if os.environ.get("FAPD_LOGPATH"):
-            self.fapd_profiling_stdout = os.environ.get("FAPD_LOGPATH") + ".stdout"
-            self.fapd_profiling_stderr = os.environ.get("FAPD_LOGPATH") + ".stderr"
+        fapd_logpath_tmp = None
```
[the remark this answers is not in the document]
tparchambault: I'll remove that assignment in line 61. An artifact of troubleshooting. Env vars have to be explicitly passed as args to the pkexec command in the /usr/sbin/ wrapper script and consequently are always set (even if to an empty string) from the perspective of fapolicy-analyzer. I'll verify that FAPD_LOGPATH works correctly in both dev and rpm installs.

fapolicy_analyzer/ui/fapd_manager.py
```diff
@@ -54,41 +58,43 @@ def __init__(self, bSuFapdControl=True):
...
+        fapd_logpath_tmp = os.environ.get("FAPD_LOGPATH")
+        if fapd_logpath_tmp and len(fapd_logpath_tmp):
```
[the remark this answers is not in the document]
tparchambault: See above wrt that env var being always set. I'll verify it works in both the dev env and when using the wrapper script w/pkexec.

fapolicy_analyzer/ui/fapd_manager.py
tparchambault: Done.

tparchambault: Will push after I address the FAPD_LOGPATH issue below.

fapolicy_analyzer/ui/fapd_manager.py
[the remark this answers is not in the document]
dorschs57:
```suggestion
        fapd_logpath_tmp = os.getenv("FAPD_LOGPATH")
        if fapd_logpath_tmp:
```
That should work.  `os.getenv` will return the default of None if the env is not available.

fapolicy_analyzer/ui/fapd_manager.py
tparchambault: Left the self.mode attribute as part of the public interface, for the client code to select the desired target fapd instance to direct start/stop/status requests. Added two new attributes to represent the current active fapd instance and the prior instance (if any). The intention is to return the online fapd instance to the same pre-profiling run state after profiling is completed.

fapolicy_analyzer/ui/fapd_manager.py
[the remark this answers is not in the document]
tparchambault: Yeah, agreed but in the pkexec runtime env that env var will always be set even if it's a Null string so None is never returned from getenv (it returns ""). It will work correctly in the dev env though. But your statement about a string with zero length = False allows things to work correctly in either dev or deployed (pkexec) env, so I used that approach. 

> since empty string is also a falsy value.

Thx for that. Hopefully it's good now.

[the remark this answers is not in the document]
tparchambault: Verified in both dev and rpm deployed environments.

fapolicy_analyzer/ui/fapd_manager.py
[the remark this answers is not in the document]
tparchambault: Resolved the issue by using the fact that an empty string evaluates to a boolean False.

fapolicy_analyzer/ui/fapd_manager.py
tparchambault: Removed all accessor functions. Went with the "Pythonic" approach of not using accessors w/simple attributes.

dorschs57 (approved):

PR #466, opened 2022-04-06
adding rule ID

### LINKED ISSUE #455: Rule ID in Guided Editor
Render the rule id to the left of the rule text in the guided editor.

egbicker:
Closes #455

### DIFF
--- a/fapolicy_analyzer/tests/rules/test_rules_list_view.py
+++ b/fapolicy_analyzer/tests/rules/test_rules_list_view.py
@@ -70,7 +70,7 @@ def test_creates_widget(widget):
 def test_renders_rules(widget):
     widget.render_rules(mock_rules)
     model = widget.get_object("treeView").get_model()
-    assert [r.text for r in mock_rules] == [x[0] for x in model]
+    assert [str(r.id) + " " + r.text for r in mock_rules] == [x[0] for x in model]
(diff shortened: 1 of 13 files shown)


jw3:
The id field is rendered as a `-1` on the info nodes

image

egbicker:
Info node ID issue is fixed

jw3:
I can't test again atm but is it using the rule I'd for the leaf?  I think they should just be implied.

egbicker:
Leaf being the left-hand column?

jw3:
Did that break the sort order?

image

egbicker:
Yeah, the first fix I pushed that broke sorting was to treat the ID like a string so that I could use an empty string display in the child row but Gtk.SortType only has Ascending and Descending

jw3:
LGTM :+1:

dorschs57:
When I have bad rule that generates an error to display I get the error below when I click to expand the child row

image

dorschs57:
The only other thing I see and it can be a later issue to clean up so this can be merged is when I expand the child rows, column 2 seems to bump over some. It causes the table data to "jump" to the right.  Again I wouldn't hold up the merge for that one.

dorschs57:
> When I have bad rule that generates an error to display I get the error below when I click to expand the child row
> 
> image

No longer seeing this issue. Not sure what was causing it but I can't reproduce it.

PR #483, opened 2022-04-20
Reverted the use of python 3.9 ThreadPoolExecutor arguments

### LINKED ISSUE #422: Update the calls to the ThreadPoolExecutor to be compatible with python 3.6
The API for the `ThreadPoolExecutor` has changed between python 3.6 and 3.9. We need to look at this and how to make our code compatible with python 3.6 but still keep the same functionality.

dorschs57:
Removed the use of the cancel_futures parameter to the ThreadPoolExecutor because it is not available in python 3.6. I've tested this and I think we are good with turning off this cancel functionality. I think it was an issue when you closed the application before the System Trust data loaded completely. The process would hang until the thread pulling the trust data completed, but this isn't an issue now because the rust layer is pre-caching this data during startup. We should be good to revert to the older python 3.6 version now.

To test this you can start up the application. Then close the UI window before the system trust is fully loaded. It should close cleanly without hanging.

closes #422

### DIFF
--- a/fapolicy_analyzer/ui/features/system_feature.py
+++ b/fapolicy_analyzer/ui/features/system_feature.py
@@ -103,7 +103,7 @@ def create_system_feature(
             _checkpoint = _system
 
             if executor:
-                executor.shutdown(cancel_futures=True)
+                executor.shutdown()
 
             if system:
                 dispatch(system_initialized())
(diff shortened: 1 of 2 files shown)


tparchambault (approved):

tparchambault:
Tested in both FC34 dev and clean FC34 deployed via rpm environments w/o any observed issue. Performed the suggested terminate while the trust dbase is still loading w/no issues nor excessive UI latencies observed.

PR #487, opened 2022-04-21
Fixed broken typing and tests in python 3.6

### LINKED ISSUE #486: Fix python typing incompatibilities with python 3.6
There were some changes in python typing between python 3.9 and 3.6 that will have to be fixed before downgrading to python 3.6.

dorschs57:
Fixed some bad typing in `rules_list_view.py` that worked in python 3.9 but not 3.6.  Also fixed some tests that break in python 3.6 because of changes in the pytest `call` objects API.

closes #486


tparchambault (approved):

tparchambault:
Verified in dev environment and locally running unit-test suite. No observable issues.

PR #493, opened 2022-05-11
Fixing CI pipeline with apt-get update, and specific docker image version

tparchambault:
Updated RHEL image version to fix rpm builds. Added `apt-get update` to refresh the rpm version info prior to installing the failing libdbus-1-dev pkg.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,7 +27,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install requirements
-        run: sudo apt-get install libdbus-1-dev
+        run: |
+          sudo apt-get update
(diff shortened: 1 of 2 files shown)


tparchambault:
Don't know why `apt-get upddate` is now critical for installing libdbus-1-dev pkg. I noticed that all test that installed that pkg failed except for `UI Test Suite for Python 3.x` which performed the `apt-get update` prior to the installation of libdbus-1-dev.

I got the docker image information from the redhat site. I don't know if they just bumped to RHEL8.6 images, so just added complete file versioning info. Don't know if this is a work-around for not being able to reach a repo server, or a path has changed, or the new release implies that older releases need complete naming supplied in the download request.

dorschs57 (approved):

PR #494, opened 2022-05-18
Add rule file markers to guided rule editor

### LINKED ISSUE #458: Show the file marker for managed editor view of rules
Update the Managed Editor rules view to support displaying the file markers from the `rules.d` directory.

dorschs57:
This adds the rule file marker rows to the table view in the Rules Admin tool's Guided Editor. I'm not sold on the color scheme here, but with adding the collapsible rows for the rule file markers a simple alternating white/light gray pattern didn't look right. 

closes #458

### DIFF
--- a/fapolicy_analyzer/tests/rules/test_rules_list_view.py
+++ b/fapolicy_analyzer/tests/rules/test_rules_list_view.py
@@ -84,8 +84,8 @@ def test_renders_origin_files(widget):
     model.foreach(get_origin_rows)
     assert sorted(set([f"[{r.origin}]" for r in mock_rules])) == [x[0] for x in origins]
     assert set([-1]) == set([x[1] for x in origins])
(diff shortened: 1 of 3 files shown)


dorschs57:
Here is my original design with alternating row colors and the darker color for the file marker rows:

image

dorschs57:
@jw3 I think this is what you are talking about.  If we decide to go with this I can see about bumping the file marker text to the left so it lines up with the rule number column.

image

jw3:
> I think this is what you are talking about.

Yes.  I like this better than the previous iteration.

> If we decide to go with this I can see about bumping the file marker text to the left so it lines up with the rule number column.

Yeah doing that could clean it up some more.

jw3 (approved):

dorschs57:
I updated the styling so the file markers line up better on the left side of the view.

image

PR #504, opened 2022-06-02
Fa362 policy to rules

egbicker:


fapolicy_analyzer/ui/main_window.py
```diff
@@ -359,10 +361,19 @@ def on_trustDbMenu_activate(self, menuitem, *args):
         self.__pack_main_content(router(ANALYZER_SELECTION.TRUST_DATABASE_ADMIN))
         self.__set_trustDbMenu_sensitive(False)
 
-    def on_rulesAdminMenu_activate(self, *args):
+    def on_rulesAdminMenu_activate_item(self, *args):
+
+        treeview = self.get_ref().get_children()[0].get_children()[0].get_children()[2].get_children()[0].get_children()[0].get_children()[2].get_children()[0].get_children()[0].get_children()[1].get_children()[0].get_children()[0].get_children()[0]
```
egbicker, line 366: This chain of get_children is inelegant and feels like there should be a better way to dig down to a specific gtk item. I have a smoother way of doing it below but this one has notably more steps to dig through

dorschs57 (changes requested):
This seems to work fine the first time I use the `Go To Rule` menu item. After the first time though I start seeing the error below. It still seems to go to the rule though. 

image

dorschs57:
Also, since a rule has both a subject and an object, shouldn't the `Go To Rule` object be present in both the subject and the object context menus from the analysis view?

PR #505, opened 2022-06-02
Update README.md

tparchambault:


jw3:
I'm not a fan of the way this renders

---

image

tparchambault:
I am not quite that dense... I assumed that. It got handled.

jw3:
How about

---

To generate rules that can be analyzed we require the following `syslog_format` configuration in `fapolicyd.conf`

`syslog_format = rule,dec,perm,uid,gid,pid,exe,:,path,ftype,trust`

PR #515, opened 2022-06-27
Add systemd control of fapd profiling instance

tparchambault:
Initial commit of fapd profiling instance w/systemd control.


jw3:
@tparchambault is this going to make it into master?

tparchambault:
> @tparchambault is this going to make it into master?

Nada. This is a working draft PR so that I could generate rpms via CI and stay in-sync w/master.

PR #524, opened 2022-06-30
Shade redux

jw3:
To prevent the RPM from installing py-redux as a separate Python package we can either
1. Have py-redux published to EPEL and consume it as a dependency
2. Move the py-redux code into our source tree (preserving the license and copyright)

Since the project is small and appears to be inactive, the second may be preferred.

The changes here
- move the py-redux core code under our root package
- remove the vendored dir and its installation in the spec and pipfile


jw3:
Need to bring in the redux tests yet.

jw3:
@dorschs57 I'd like to merge this.  Could you can take one more look, at the tests that were added.

dorschs57 (approved):
@jw3 I'm not sure we need to bring in `tests/redux/counter` and `tests/redux/todo`. Those are really just examples of how to use the library. Without them I'm not sure we would get any test coverage on the redux code though, so maybe for this round we merge them in and revisit with some real unit tests later?  I'm good with or without them in there, so I'll approve this one.

PR #526, opened 2022-07-01
Add rules.d location error text

tparchambault:
Updated start-up failure dialog text to include the new rules.d directory location error message.


dorschs57 (changes requested):
You need to run the `extract_messages` to update the template pot file with the new message.

dorschs57 (approved):

jw3 (approved):

PR #533, opened 2022-07-11
Write rules.d structure

### LINKED ISSUE #532: Write rules.d structure
Write the rules db out based on origin to produce the rules.d structure.

jw3:
Write the rules db to disk while separating files based on origin, placing the files under the configured rules.d directory. This is not required for functionality, as the compiled.rules file is already being written.  This instead being done to keep a compliant fapolicyd configuration to ensure the standard fapolicyd workflows will continue to work. 

A side effect of this is that the preexisting rules files under rules.d will be deleted prior to writing the structure from the current rules db.. The original rules files are captured in the predeploy snapshot and can be reverted if needed.

Closes #532

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -290,6 +290,7 @@ version = "0.4.0"
 dependencies = [
  "nom",
  "serde",
+ "tempfile",
  "thiserror",
 ]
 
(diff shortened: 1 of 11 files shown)


jw3:
Need to clear the .d files prior to deploy.

PR #534, opened 2022-07-12
Rule text write

dorschs57:


jw3 (approved):
Looks good.  Discovered #550 while testing, but it can be addressed separately.

dorschs57:
I'm was going to merge this PR, but saw @jw3 was using it for testing #544. Is it ok to merge this one or do we think its somehow linked ot the unresponsiveness issue?

jw3:
This is good to merge

Issue #495, opened 2022-05-19
Enable rule editing in Text View


dorschs57:
Enable editing in the Rule Admin tool's Text View so the user can edit rules and deploy the results.

jw3:
I pictured the flow being something like

- User edits text in editor
- User triggers a "compile" action
- Rules are "compiled" using a `RuleChangeset`
- On the `RuleChangeset` you can `set()` the full rule text, and then `get()` the parsed rules that have been populated with `RuleInfo` metadata from the parse/lint.  The changeset does not have to be added to a `System` to do these incremental checks
- Continue along with this process until some a "save" action occurs, at that time the `RuleChangeset` gets applied to the `System`
- User can now deploy the `System`, or more edits can be applied using additional changesets

jw3:
Ok we can stash and access the changeset source string

PR #535, opened 2022-07-13
Refine Rule Change API

jw3:
Several refinements and additions to the rule change api

- Adds a `parse(str)` method to the RuleChangeset to expose marker parsing errors
- Adds a new accessor for clarity, `rules()` to replace `get()`.
    - `get()` should be considered deprecated
- Retains the text from the previous successful parse and provides accessor `text()`
- Updates the `change_rules.py` example

The changes to the API should be non-breaking, `get()` should be removed either in this PR or after.

#495

### DIFF
--- a/crates/pyo3/src/rules.rs
+++ b/crates/pyo3/src/rules.rs
@@ -145,18 +145,22 @@ impl PyChangeset {
         rules_to_vec(self.rs.get())
     }
 
-    pub fn set(&mut self, text: String) -> bool {
-        self.rs.set(&text).is_ok()
+    pub fn set(&mut self, text: &str) -> bool {
+        self.rs.set(&text.trim()).is_ok()
     }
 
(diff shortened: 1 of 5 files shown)


jw3:
@dorschs57 do you see any remaining gaps?

dorschs57:
> @dorschs57 do you see any remaining gaps, or have comments about the API?

The only other thing I haven't come up with a plan for yet is how to display the changes to the user in the deployment confirmation dialog. With trust we show a 2 column table with each row being a file path and the `add/delete` action type. I'm not sure what to do with rules since all we have is a text blob. Any thoughts on that one? Can we indicate the changes somehow in the `vec` that comes back from the rule changeset `get` method?

jw3:
Yea, good one.  Ill look into a diff of the rule model.  Not sure exactly where it lives, but it wont be on the changeset.  Will be follow up PR.

jw3:
@dorschs57 how about adding an issue for that

PR #539, opened 2022-07-18
Rule change diff

### LINKED ISSUE #538: Bindings to rule change summarization
We need to summarize and display rule changes at deploy time, to do that we need to create a diff-like report.

It is not entirely clear how to do this, but some thoughts

1. Given that an insertion would cascade ID changes to all rules that follow, could ID-only changes be ignored like whitespace in a git diff?
2. Would a textual diff be a good first iteration?

jw3:
Provide a diff formatted summarization of changes between two rule databases.  Taking the approach of diffing the text as it appears in the editor.  This provides diffs in our application context, including file markers, and avoids rule number diffs.

This commit adds a function to the bindings with the signature of `rules_difference(System, System)` which allows you to pass in two systems and receive a string similar to the `System#rules_text()`.  The diff adds prefixes of `+` and `-` to all changed lines. 

Closes #538

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -282,6 +282,7 @@ dependencies = [
  "fapolicy-rules",
  "fapolicy-trust",
  "pyo3",
+ "similar",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


jw3:
The output of the updated example in this PR is 

```
diffing additional changes
 [05-foo.rules]
-%foo=bar,baz
+%foo=fizz,baz
 allow perm=exec all : all
 
-[10-bar.rules]
+[10-fizz.rules]
 %bing=bam,boom
 deny perm=any all : all
```

jw3:
It is up to the caller to format the output, for basic uses it is trivial as the example shows

image

dorschs57 (approved):

PR #541, opened 2022-07-20
updated test imports

dorschs57:
I switched all relative imports over to absolute imports across the entire UI code base. Long term I think this is our best move as relative imports caused issues when we started creating submodules under the `ui` module. It also is part of the fix to get our text coverage close to matching up with what I'm seeing locally. Right now CI is at 95.43% and locally I'm getting 95.73%.  A couple of caveats:

1. I had to skip the test for internationalization. I can't seem to get that working out in CI. I'm guessing its an difference between RHEL and Ubuntu OS's.
2. @jw3 I'm no longer using the binding wheel file we build in CI for the tests. I'm not sure if we want to keep building that as part of the CI or not. I feel like there is value in doing the build as part of the CI.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,10 +1,13 @@
 [run]
 source =
-    ui
-    util
+    fapolicy_analyzer/ui
+    fapolicy_analyzer/util
+omit =
+    */tests/*
+    fapolicy_analyzer/css/*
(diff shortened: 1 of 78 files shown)


jw3 (approved):
:+1:

Tested it on el8, no issues.

PR #548, opened 2022-07-24
Full daemon reload

### LINKED ISSUE #544: UI unresponsive after deployment rollback
### To reproduce

1. Add trust changes
2. Deploy
3. Select "No" when asked to keep changes

### Noteworthy

- Usually on the first or second deployment the UI becomes unresponsive.
- Same behavior whether the timer expires or you manually click No.
- Did not encounter issues If the deployment is accepted.
- The rollback of configuration is indeed written to disk (ie. the original is restored).
- The daemon is in active state after the app locks, so it appears to be properly started
- Consistently reproducible on fc34 using `rule-text-write` branch's RPM
- **Did not observe issue on RHEL 8** using the same branch

### LINKED ISSUE #547: Rule deployment requires full daemon reload
Currently on deployment we only signal the fifo pipe to reload trust.

jw3:
The reload for trust used a signal through a fifo pipe to have fapolicyd reload the trust db.  To support rule deployments we need to reload the systemd service.  In all cases we will no bring down the daemon to update configuration as there is no way we can guarantee that the write to disk would be successful while fapolicyd is up.

Closes #547
Closes #544

### DIFF
--- a/crates/daemon/src/fapolicyd.rs
+++ b/crates/daemon/src/fapolicyd.rs
@@ -8,6 +8,7 @@
 
 // todo;; tracking the fapolicyd specific bits in here to determine if bindings are worthwhile
 
+use std::io::Write;
 use std::thread::sleep;
 use std::time::Duration;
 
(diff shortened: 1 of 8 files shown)


jw3:
Needs rebased on #534

jw3:
Contains fix for the rollback portion of #544

crates/daemon/src/fapolicyd.rs
```diff
@@ -32,7 +35,7 @@ pub enum Version {
 }
 
 /// send signal to fapolicyd FIFO pipe to reload the trust database
-pub fn reload_databases() -> Result<(), Error> {
+pub fn signal_trust_reload() -> Result<(), Error> {
```
jw3, line 38: There is no value in keeping this.  We cannot assume to write anything to disk without first taking the daemon down. If the daemon comes down there is no need for signal.

PR #561, opened 2022-07-26
Add rules validation to rule text edit

### LINKED ISSUE #559: Enable rule text validation
Give the users the ability to validate rules before saving. Also provide the option to save first, but have it cancel out with a message if the rules are not valid for saving.

dorschs57:
Adds a new validate button to the toolbar to validate rule changes prior to saving. If there are errors or warnings a toaster message is show and the Status Information section is updated. The Save button was also update to not allow saving if there are invalid rules in the text.

closes #559

### DIFF
--- a/fapolicy_analyzer/locale/fapolicy_analyzer.pot
+++ b/fapolicy_analyzer/locale/fapolicy_analyzer.pot
@@ -7,9 +7,9 @@
 #, fuzzy
 msgid ""
 msgstr ""
-"Project-Id-Version: fapolicy-analyzer 0.3.1+101.ga756fb6.dirty\n"
+"Project-Id-Version: fapolicy-analyzer 0.3.1+104.g7857488.dirty\n"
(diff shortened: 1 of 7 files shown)


jw3 (commented):
It seems that the deploy button does not disable once enabled.

1. Save valid rules
2. Observe the deploy button becomes active
3. Add and save an invalid rule
4. Observe the deploy button remains active

Neither the validate or save button change behavior of 3-4

dorschs57:
> It seems that the deploy button does not disable once enabled.
> 
> 1. Save valid rules
> 2. Observe the deploy button becomes active
> 3. Add and save an invalid rule
> 4. Observe the deploy button remains active
> 
> Neither the validate or save button change behavior of 3-4

@jw3 Looking at your steps the deploy button would enable with the first successful change of a rules changeset. The 2nd save never applies the changeset, so the list of changesets to be deployed is still valid. If you clicked deploy after step 3, you should only be deploying the original valid rule set.

jw3 (approved):

PR #564, opened 2022-07-27
Renamed and reordered tabs in Rules Admin tool

dorschs57:
Reordered the tabs so the rules text editor was first. Also renamed the tabs to "Rules Editor" and "Rules View".


jw3 (approved):
:+1:

PR #560, opened 2022-07-26
Add exception handling around fapd handle is_active() call

tparchambault:

PR #566, opened 2022-07-28
Active checks with dbus

### LINKED ISSUE #565: Service status checks through dbus
Implement the service active checks through dbus to avoid the potentially blocked systemctl call.

The lack of dbus support for this complicates several areas, and contributes to #563.

jw3:
Eliminate the calls to systemctl for service status checks, use dbus instead.  This removes potential pinch point where deployed rules can limit the call to systemctl.

Also updates to align both stages of deployment.  Those stages, deployment and rollback, used to be different in that deployment only was a pipe write, while deployment was a full daemon reload.  Then the pipe write went away, but there were some straggling issues that were left.  This should resolve.

This also fixes the monitoring while fapolicyd is deployed with deny any all all rules.

Closes #565

### DIFF
--- a/crates/daemon/src/error.rs
+++ b/crates/daemon/src/error.rs
@@ -23,7 +23,4 @@ pub enum Error {
 
     #[error("{0}")]
     ServiceCheckFailure(String),
-
-    #[error("Failed to parse systemctl output")]
-    ServiceCheckParseFailure(#[from] FromUtf8Error),
 }
--- a/crates/daemon/src/error.rs
+++ b/crates/daemon/src/error.rs
@@ -6,7 +6,6 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
-use std::string::FromUtf8Error;
 use thiserror::Error;
 
 // errors that can occur in this crate
--- a/crates/daemon/src/fapolicyd.rs
+++ b/crates/daemon/src/fapolicyd.rs
@@ -8,13 +8,6 @@
 
 // todo;; tracking the fapolicyd specific bits in here to determine if bindings are worthwhile
 
-use std::thread::sleep;
-use std::time::Duration;
-
-use crate::error::Error;
-use crate::error::Error::FapolicydReloadFail;
-use crate::svc::Handle;
-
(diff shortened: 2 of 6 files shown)


jw3:
@tparchambault see how these changes play with your testing

jw3:
@tparchambault eh hold off, there is another item to address

tparchambault:
Since you committed some of your work 12 hrs ago and I put some observations into that other PR, did you see the following wrt timing, and you intend for this change to also address it. I believe it will.

https://github.com/ctc-oss/fapolicy-analyzer/pull/560#issuecomment-1197519179

jw3:
Yep, this PR will address the timing issues too.

jw3:
@tparchambault   OK,  I just tested the latest on rhel, `fapolicy-analyzer-0.0.0-1.255.ge34e2ed.el8.x86_64.rpm`

See what you see with it.

tparchambault:
Clean FC34 with only emacs, fapolicyd, and the generated fapolicy-analyzer rpm associated with this PR. Things look good. Ordering of fapolicyd shutdown/start reported in the journal is correct, deployed changes to the ancillary dbase were good, deployment timeout rollback also good.

tparchambault (approved):
I'm assuming the pyo3 daemon.rs deploy() throws exceptions lower down the stack on errors, i.e. from the calls it invokes and their children? I only saw this `PyRuntimeError::new_err("Daemon unresponsive")` in `fn wait_for_daemon(target_state: State)`

I'll dig around but for the sake of merging, if the answer is `yes`, we are good to go.

jw3:
@tparchambault

> Daemon unresponsive

This was an indeed an issue.  When fapolicyd was in a `failed` state (eg timed out on the max secs), `wait_for_daemon` didnt consider `Failed` equivalent to `Inactive`.  This is now fixed.

Take another look and see if it all still functions as expected.

jw3:
I am going to leave the stderr prints in there for the near term.

tparchambault (approved):

tparchambault:
Functionally, on a clean FC34 rpm install, no issues observed.

PR #567, opened 2022-07-28
Preload resources and handle errors

### LINKED ISSUE #562: Ensure deployment dialogs are in memory
We need to avoid having to load the glade files for deployment dialogs from disk.  It should be preloaded and retained in memory.

```
GLib.Error: g-file-error-quark: Failed to open file “/usr/lib/python3.6/site-packages/fapolicy_analyzer/glade/deploy_confirm_dialog.glade”: Operation not permitted (22)
```

1. Deploy a deny any all all rule
2. Stop the daemon with the System menu
3. Restart the app
4. Start the daemon with the System menu
5. Try to deploy

At step 5 you should observe the error above.

dorschs57:
Preloads system resource files like glade and CSS files into memory when it can. For files that can't be preloaded, like the SourceView language file, the exceptions are caught and the application will run without the file.

closes #562


jw3 (approved):
:+1: Looks good after testing several scenarios that previously failed.

PR #577, opened 2022-08-01
More rule lints

jw3:
- Path existence checks for all Object types
- Duplicate rules detection

### DIFF
--- a/crates/rules/src/linter/findings.rs
+++ b/crates/rules/src/linter/findings.rs
@@ -7,6 +7,7 @@
  */
 
 use crate::db::DB;
+use crate::object::Part;
 use crate::Rule;
 use std::path::PathBuf;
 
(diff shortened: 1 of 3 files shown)


dorschs57:
How do I trigger the duplicate rule warning?  I have 
```
1 deny perm=any all : dir=/home/addorschs/fapolicy-analyzer/fapolicy_analyzer/glade/ 
2 deny perm=any all : dir=/home/addorschs/fapolicy-analyzer/fapolicy_analyzer/glade/ 
```
in my rules file, but running `examples/show_rules.py` does not display that warning.

dorschs57:
Also how do you trigger the object is a wrong type warning? I have
```
3 deny perm=any all : dir=/home/addorschs/fapolicy-analyzer/fapolicy_analyzer/css/notification.css 
```
Which I expected to show that warning because the `dir` points to a file, but it did not.

jw3:
@dorschs57 All three of those work from the app and the example for me.

PR #579, opened 2022-08-02
Profiler gui updates

egbicker:


fapolicy_analyzer/tests/test_profiler_page.py
```diff
@@ -13,22 +13,28 @@
...
-        }
-        return entryDict
+def test_creates_widget(widget):
+    assert type(widget.get_ref()) is Gtk.Box
+
+
+def test_run_analyzer(widget):
+    widget._fapd_profiler = FaProfiler(FapdManager(False))
+    widget.get_object("dirEntry").set_text("/tmp")
+    widget.on_test_activate()
```
dorschs57, line 40: Is there an assertion you can use to verify the profiler is activated?

fapolicy_analyzer/ui/main_window.py
```diff
@@ -381,29 +383,12 @@ def on_rulesAdminMenu_activate(self, *args, **kwargs):
...
-
-        # Add termination and/or close button
-        dlgProfTest.get_ref().destroy()
-
-        self.__pack_main_content(
-            router(ANALYZER_SELECTION.ANALYZE_FROM_AUDIT, fapd_prof_stderr)
-        )
+        page = router(ANALYZER_SELECTION.PROFILER)
+        width = self.get_object("mainWindow").get_size()[0]
+        page.get_object("dirEntry").set_property("width_request", int(width * Sizing.PROFILER_TEXT_ENTRY))
```
dorschs57, line 388: It would be nice to handle this sizing internally in the `ProfilerPage` class. There might be a signal you can use that fires once the widget is packed into the content area where you can resize it. Having this code in `main_window` just seems out of place.

fapolicy_analyzer/ui/main_window.py
```diff
@@ -381,29 +383,12 @@ def on_rulesAdminMenu_activate(self, *args, **kwargs):
...
+        page._fapd_mgr = self._fapd_mgr
+        page._fapd_profiler = self._fapd_profiler
```
dorschs57, line 390: These properties should be set with public setters instead of access the "protected" class properties. The other option would be to accept these as params to the `__init__` of `ProfilerPage`. The `router` gives the option of passing additional data to the `__inti__` method of the route. That is probably a cleaner way to handle this.

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -0,0 +1,109 @@
...
+        self._fapd_profiler = None
+        self.running = False
+
+    def stop_button_sensitivity(self):
+        return self.running
+
+    def start_button_sensitivity(self):
+        return not self.running
+
+    def get_text(self):
```
dorschs57, line 66: Maybe rename this method to somethin like `get_entries` that indicates you are returning something more than text. `get_text` makes me think I should expect a string back. You can also use typing to indicate what you are returning 👍

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -0,0 +1,109 @@
...
+        if work_dir is not None:
+            files = glob.glob(work_dir + "/*")
+            latest_file = max(files, key=os.path.getctime)
+            time = "_".join(latest_file.split("_")[2:5])
+            run_files = [f for f in files if time in f]
+
+            for run_file in run_files:
+                buff.insert_markup(buff.get_end_iter(), f"<b>{run_file}</b>", -1)
+                with open(run_file, "r") as f:
+                    lines = f.readlines()
```
dorschs57, line 90: You need to make sure you handle any exceptions during the read here.  Think of @jw3 testing with `deny perm=any all : all`.

fapolicy_analyzer/ui/main_window.py
dorschs57: Also, the `self._fapd_profiler` is created here and only used to pass into the profiler page. I think you can just move its creation right into the `__init__` of `ProfilerPage` and eliminate it as a parameter that is passed in.

dorschs57 (changes requested):

fapolicy_analyzer/ui/main_window.py
egbicker: I didn't want to make a second instance of the _fapd_manager inside of main_window, but I agree that passing it in through the router is a cleaner option

dorschs57: Yeah I agree about not making a 2nd copy of `_fapd_manager`, but `_fapd_profiler` is only used in `main_window` to pass it thru to the profiler page. Why not just instantiate the `_fapd_profiler` there?

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        UIPage.__init__(self, actions)
+        manager, window = data
+        self._fapd_profiler = FaProfiler(manager)
+        self.running = False
+
+        width = window.get_size()[0]
+        self.get_object("dirEntry").set_property("width_request", int(width * Sizing.PROFILER_TEXT_ENTRY))
+
```
dorschs57, line 61: Instead of passing in the window object, it looks like you can wire up the `realize` signal. That one fires whenever the widget is added to a window. A handler like this will then do your resizing with no extra params needed.

```
    def on_profilerPage_realize(self, widget):
        window = widget.get_toplevel()
        width = window.get_size()[0]
        self.get_object("dirEntry").set_property(
            "width_request", int(width * Sizing.PROFILER_TEXT_ENTRY)
        )
```

dorschs57: Also this resizing doesn't hold up if I resize my screen once the profiler page is initially rendered. You'll probably have to wire up to another signal for that one. There is probably some resized signal to listen for.

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -0,0 +1,114 @@
...
+import os
+
+from time import sleep
+from fapolicy_analyzer.ui.store import get_system_feature
+from fapolicy_analyzer.ui.ui_page import UIAction, UIPage
+from fapolicy_analyzer.ui.ui_widget import UIConnectedWidget
+from fapolicy_analyzer.ui.configs import Sizing
+from fapolicy_analyzer.ui.faprofiler import FaProfiler
+gi.require_version("Gtk", "3.0")
+from gi.repository import Gtk
```
dorschs57, line 27:
```suggestion
gi.require_version("Gtk", "3.0")
from gi.repository import Gtk   # isort: skip
```

Can you please add the isort comment? I use a plugin that sorts my imports and that prevents it from being sorted above the GTK version check line.

fapolicy_analyzer/ui/profiler_page.py
egbicker: The main issue is that the `get_toplevel` doesn't return the `Window` object from `main_window` when called from `profiler_page`, it returns a `Box` object.

dorschs57: With the code above and using the `realize` signal, my `get_toplevel` returns the same object as the `window` passed into `__init__`. I think once the widget is added and the `realize` signal is fired, the toplevel points to the window, before that it points to the widgets "toplevel" which is a box because that's all it knows about.

dorschs57: Fixed with my last commit

dorschs57 (approved):

jw3:
Looking good.  I still need to pull the new RPM with the button wired up yet, but the execution otherwise seemed right.

tparchambault:
Almost all good: All base functionality is working except the Run Analysis button is not presenting properly with only a toaster msg showing: "An error occurred trying to parse the event log file. Please try again or select a different file."

On a second run, manually transitioned to the Analysis tool view and loaded the fapd*.stderr and observed no parsing nor display issues.

egbicker:
I think this latest commit solved @tparchambault issue if you want to give it another look @jw3

jw3:
It worked the first time, but now I get

image

jw3:
Is the analyze button going to move to the toolbar?

egbicker:
That is not my current plan. The items on the toolbar will persist across the multi-target tabs we have planned so I thought it made more sense to have an analyze button for each tab that would correspond with that target file.

egbicker:
Does that lead to a relevant failure rate where the user is on the wrong tab when they hit the button? Or is the simplified UI better?

jw3:
I'd think it would be pretty clear what was happening, that is the run you have selected is the run getting analyzed.

It will be especially clear right now since we only support a single run.

What else is blocking the merge of this?

egbicker:
Nothing else. Was just waiting on your testing to see if it was good to go. I can move the analyze button if that's the last hangup.

jw3:
Yeah, I'd say move it and then its gtg

jw3:
Is there an icon for this?

image

egbicker:
image

jw3:
I'm guessing that's not rhel 8

egbicker:
Correct. Ubuntu 20.04

jw3:
Set the icon to `applications-science`

image

HT `gtk3-icon-browser`

Issue #531, opened 2022-07-08
Detect systemd service start/stop events to replace hard coded delays


tparchambault:
The original implementation used conservative hard coded delays to sync fapolicyd starts/stops {via dbus} with the invocation and termination of the Profiling target executables. The focus at that time was correct functionality however it's time for a quick first pass optimization to improve the end-user experience. We can totally eliminate or minimize the hard coded delays if we wait/block while monitoring systemd's journal for fapolicyd events of interest e.g. wait after starting the fapolicyd profiling instance until journalctl reports that the fapd_profiling.service has started, then invoke the profiling target executable. 

This is not a blocker to the systemd profiler work, and will be done in parallel.

jw3:
This should be able to be done on the backend with a few aditions.

1. Could listen for dbus events
  - would look similar to https://github.com/diwic/dbus-rs/blob/master/dbus/examples/unity_focused_window.rs
2. Register a Python callback with the backend that is called when state changes

We don't have any examples of 2 yet but it should be straightforward.  I'll take a look at this in milestone 8.

Issue #508, opened 2022-06-10
Add background activity spinner widget when a profiling session is executing


tparchambault:
Add a spinner in the foreground above the modal profiling execution dialog.

tparchambault:
The current release provides appropriate real-time feedback to the user during a profiling test. The output text UI element displays current profiler state information during target execution and event log analysis.

This ticket will be closed.

PR #597, opened 2022-08-16
Profiler backend

jw3:
## Experimental profiler backend

This is a first pass at backend functionality with initial bindings.  Future PR(s) will expand the bindings and integrate with the Python `faprofiler`.

### todo
- [x] does the profiler live in its own crate?
- [x] review the `wait_for_daemon` functions in daemon and pyo3 crates and eliminate one if possible
- [x] add remaining clap arguments 

### Goals
1. Provide Rust backend that matches the profiling APIs defined by `fapd_manager.py` and `faprofiler.py`
2. Shift the profiling flow control backwards into Rust.
3. Provide a standalone executable that can profile without the GUI
4. Move  the profiler control into systemd
5. Stateless lock-free control
6. Provide bindings that allow the UI to be unchanged 
7. Allow non-deployed rules to be profiled

### Accomplished
1. > Provide Rust backend that matches the profiling APIs defined by `fapd_manager.py` and `faprofiler.py`
    - Rust backend :heavy_check_mark: 
    - Matching APIs - Follow up PRs
2. > Shift the profiling flow control backwards into Rust.
    - `daemon::Profiler` :heavy_check_mark: 
3. > Provide a standalone executable that can profile without the GUI
    - Added `faprofiler` as a CLI tool :heavy_check_mark: 
4. > Move  the profiler control into systemd
    - :heavy_check_mark: 
6. > Stateless lock-free control
    - Stateless: there is state to maintain when supporting optional features
    - lock-free: client concern
8. > Provide bindings that allow the UI to be unchanged
    - Follow up PRs
9. > Allow non-deployed rules to be profiled
    - CLI :heavy_check_mark:
    - GUI: will need some additional support

### Impacts
- #508 
- #531

### todo
- Determine if the target should be a systemd unit

### DIFF
--- a/.github/workflows/tools.yml
+++ b/.github/workflows/tools.yml
@@ -52,8 +52,11 @@ jobs:
       - name: Export files
         run: |
           mkdir /tmp/tools
-          mv target/release/tdb /tmp/tools/tdb-${{ steps.tag_name.outputs.VERSION }}
-          mv target/release/rulec /tmp/tools/rulec-${{ steps.tag_name.outputs.VERSION }}
+          mv target/release/tdb /tmp/tools/tdb-${APP_VERSION}
+          mv target/release/rulec /tmp/tools/rulec-${APP_VERSION}
+          mv target/release/faprofiler /tmp/tools/faprofiler-${APP_VERSION}
+        env:
+          APP_VERSION: ${{ steps.tag_name.outputs.VERSION }}
 
       - name: Archive Tools
         uses: actions/upload-artifact@v2
--- a/.github/workflows/tools.yml
+++ b/.github/workflows/tools.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install build deps
         run: |
           dnf groupinstall -y "Development Tools"
-          dnf install -y git dbus-devel
+          dnf install -y dbus-devel
 
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
@@ -31,7 +31,7 @@ jobs:
 
       - uses: actions/checkout@v2
         with:
-          ref: ${{ github.head_ref }}
+          ref: ${{ github.ref }}
           fetch-depth: 0
 
       - name: Build tools
--- a/.github/workflows/tools.yml
+++ b/.github/workflows/tools.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Install build deps
         run: |
           dnf groupinstall -y "Development Tools"
-          dnf install -y dbus-devel
+          dnf install -y git dbus-devel
 
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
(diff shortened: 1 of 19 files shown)


jw3:
This flow currently works through an on-demand system drop-in unit

```rust
    let mut profiler = Profiler::new();
    profiler.activate()?;
    let out = Command::new("ls").arg("/tmp").output()?;
    println!("{}", String::from_utf8(out.stdout)?);
    profiler.deactivate()?;
```

tparchambault:
I'm good with your goals and your approach. Two comments

> Stateless lock-free control

We have to handle system state at some level such that two fapd instances are not running simultaneously. If we are using service units maybe the `Conflicts=xxx` directive will address this.

> Determine if the target should be a systemd unit

Even if we containerize the stand-alone, there's no reason that it can't be controlled as a systemd service.

Re: `drop-in` terminology:
> ... an on-demand system drop-in unit

We may have differing ideas of `drop-in` and my idea of it may be out-dated because I'm looking at manpages associated with an older systemd release, and systemd may have 'evolved.'

I took a quick look at your code and got the idea that you are dynamically generating the service unit file. No issues with that approach btw.

I'm trying to use the systemd meaning of the term although it's not particularly intuitive to my own personal idea of what a `drop-in` would be, but they do explicitly specify the mechanism. In any event, from `man systemd.unit`:

>  Along with a unit file foo.service, a "drop-in" directory
       foo.service.d/ may exist. All files with the suffix ".conf" from this
       directory will be parsed after the file itself is parsed. This is
       useful to alter or add configuration settings for a unit, without
       having to modify unit files. Each drop-in file must have appropriate
       section headers. Note that for instantiated units, this logic will
       first look for the instance ".d/" subdirectory and read its ".conf"
       files, followed by the template ".d/" subdirectory and the ".conf"
       files there.

So basically magic sub-directories for overrides. We could use them to override and/or extend the declarations in the shipped fapolicyd.service file to specify different stdout/stderr locations, add the --debug and --permission CL args, etc, but I'm not sure that's what you were thinking wrt `drop-ins` _or_ I could have missed something and that's exactly what you were thinking.


I'm good with all of your ideas. I like consistent APIs and control flow between similar functional sub-systems. I'll look at the code in more detail later today.

jw3:
>> Stateless lock-free control

> We have to handle system state at some level such that two fapd instances are not running simultaneously. If we are using service units maybe the `Conflicts=xxx` directive will address this.

The goal is that checking `is_active` on a handle is enough.

>> Determine if the target should be a systemd unit

> Even if we containerize the stand-alone, there's no reason that it can't be controlled as a systemd service.

Concur.  The initial impl here is going to remain Console, but that could be changed to a `one-shot` unit that gets cleaned up at the end of the run.

> Re: `drop-in` terminology:

Your definition is accurate.  My original intent was instead of writing a service file to add drop-ins on fapolicyd, but writing a separate service file ( `fapolicyp` :stuck_out_tongue: )  is a more reliable solution to determine what is running on the system without relying on maintaining state in the application.

> I'm good with all of your ideas.

:+1: 

Now that 0.6.0 is branched off this is going to get cleaned up and merged to master.  It doesn't have to be the backend yet, but I don't want it to get stale.  I also want the cli tool to play with regularly.

PR #599, opened 2022-08-17
Fixed falsy checks in rules analysis page

### LINKED ISSUE #592: Analysis does not show events with uid 0
Events with `uid=0` do not load up subject or object columns in the analysis.

Not sure where the breakdown is here.  Some testing with the `examples/parse_events.py` script is successful, so it appears it might be on the UI side of the bindings.

Test data

```
rule=1 dec=deny perm=open uid=0 gid=0 pid=55523 exe=/usr/bin/ls : path=/etc/ld.so.cache ftype=application/octet-stream trust=0
rule=1 dec=deny perm=open uid=1 gid=0 pid=55523 exe=/usr/bin/ls : path=/etc/ld.so.cache ftype=application/octet-stream trust=0
```

Analysis will show all details of the event of `uid=1` but not `uid=0`.

dorschs57:
Falsy checks where to blame for this one. It should be good now.

closes #592

### DIFF
--- a/fapolicy_analyzer/tests/mocks.py
+++ b/fapolicy_analyzer/tests/mocks.py
@@ -26,17 +26,17 @@ def mock_trust():
 
 
 def mock_users():
-    mockUser1 = MagicMock(id=1)
+    mockUser1 = MagicMock(id=0)
     mockUser1.name = "fooUser"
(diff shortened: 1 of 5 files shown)


jw3 (approved):
:+1:

PR #609, opened 2022-08-18
Change post deployment revert so changes are kept in the UI

### LINKED ISSUE #550: Rolling back deployment discards changes
When rolling back a deployment the original system should be written to disk, but the application should continue to function with the modified system.

1. Change some rules
2. Deploy
3. Select No to rollback changes
4. Observe that your edits are lost

I would want to deploy and test my changes, but if I rollback I want to start editing where I left off.  The editor should facilitate a rapid rules development loop.  Write, Validate, Run, Modify, Validate, Run, ...

dorschs57:
Changed the UI behavior so changes are kept in the UI if the user decides to rollback the deploy when the post-deployment revert dialog. The functionality of loading a saved session file should be the same. When loading the file, current changes are cleared and the set of changes from the session file are loaded.

closes #550

### DIFF
--- a/fapolicy_analyzer/tests/features/test_system_feature.py
+++ b/fapolicy_analyzer/tests/features/test_system_feature.py
@@ -20,12 +20,13 @@ import fapolicy_analyzer.ui.store as store
 import pytest
 from callee import InstanceOf
 from callee.attributes import Attrs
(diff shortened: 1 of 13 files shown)


jw3 (approved):
:+1:

PR #612, opened 2022-08-18
Add blocking fapd Handle.wait_until_active/inactive() calls

tparchambault:
Phasing in explicit daemon query/waits to replace magic hard coded delays to wait for the on-line fapolicyd instance to come up or shut down when transitioning between on-line active service and FA profiling in permissive mode.

Need to create something functionally equivalent for the subprocess.Popen() fapolicyd profiling instance monitoring until it is migrated to a systemd service.


jw3:
@tparchambault Is this ready for review?

PR #613, opened 2022-08-18
Updated toaster notifications to handle multiple notifications

### LINKED ISSUE #584: Add toaster message timeouts
Currently the toaster messages have to be dismissed by the user. This can cause toasters to stack up behind one another. We should probably have some timeout so they dismiss themselves after so long. Maybe error message still have to be physically dismissed by the user.  

It would also be nice if toasters scrolled down the screen if there are multiple messages instead of stacking up behind one another.

dorschs57:
Updated the toaster notification so they will scroll down the screen for the first 5 notifications. After 5 new ones are not displayed until some of the currently displayed ones are dismissed. Also update WARN notification to timeout after 10 seconds. Now only ERROR notifications must be dismissed by the user.

closes #584


dorschs57:
@jw3 give this branch a try. It might help out with showing multiple rule validation failures from the editor page.

jw3 (commented):
1. There is some odd resizing behavior when closing out toasters with small messages to reveal toasters with larger messages.
2. I like the timeout.

Not sure how I feel about the stacking in general.  I might like a destructive toaster; one  that dismisses the previous message... but that raises other issues.  Going to let this one simmer for now.

Android toasters overwrite the previous, that's probably where that thought is coming from.

jw3:
I'm going to close this for now, since its stale and conflicted from picking part of it out already.

The original issue #584 is open and tracks the original need.

Issue #536, opened 2022-07-13
Defining Python Exceptions


jw3:
Custom exceptions are needed to convey context of failure.

https://pyo3.rs/main/exception.html

Appears that #517 may be impacting the visibility of these from Python.

#459 
#535
#623

PR #623, opened 2022-08-22
Prevent empty rules

### LINKED ISSUE #604: System should prevent deploying empty rule set

jw3:
Detect empty rules and throw a parse error to prevent the system from deploying and ultimately blowing up the fapolicyd reload.

Classifying the empty-rule failure in same bucket as a malformed-marker parse, doing that stops the changeset from being populated which seems to be the right spot in the flow to put a halt to things.

This impacts the bindings by dropping the `set` call which never throws, and forcing the use of the `parse` call that may throw.  Forcing the client to be aware of the potential parse error here is important.

Closes #604

### DIFF
--- a/crates/pyo3/src/rules.rs
+++ b/crates/pyo3/src/rules.rs
@@ -11,7 +11,7 @@ use pyo3::{exceptions, PyObjectProtocol};
 
 use fapolicy_rules::db::Entry::*;
 use fapolicy_rules::db::{Entry, DB};
-use fapolicy_rules::error::Error::MalformedFileMarker;
+use fapolicy_rules::error::Error::{MalformedFileMarker, ZeroRulesDefined};
 use fapolicy_rules::ops::Changeset;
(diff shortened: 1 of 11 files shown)


jw3:
@dorschs57 I sketched in some debug logs at the impacted callsite for this.  Can you review the changeset wrapper change (dropping the non-throwing set call) and update the other places with the logging placeholders?

dorschs57:
> @dorschs57 I sketched in some debug logs at the impacted callsite for this. Can you review the changeset wrapper change (dropping the non-throwing set call) and update the other places with the logging placeholders?

@jw3 take a look at my last commit and see if that is what you were thinking.

jw3:
Yeah, that looks good.  There are two user induced failures possible at the point of this parse

1. Malformed file marker
2. Zero rules defined

There could also be IO or other non-user related issues.

We need to reflect the cause of the failure in the toaster, so we really need #536 to do this right.  Until then the log details will work.

PR #629, opened 2022-08-25
Preserve profiler config

### LINKED ISSUE #605: Restore parameters from last run on profiler UI
When switching to the profiler can we populate the fields with the last run.

This state does not need preserved after the lifetime of the application.

egbicker:
Preserve the configuration in the profiler config fields while switching back and forth between the other gui screen.

Closes #605

### DIFF
--- a/fapolicy_analyzer/ui/profiler_page.py
+++ b/fapolicy_analyzer/ui/profiler_page.py
@@ -95,6 +95,7 @@ class ProfilerPage(UIConnectedWidget, UIPage, Events):
         text = buff.get_text(buff.get_start_iter(), buff. get_end_iter(), True)
         if not text == system.get("profiler").output:
             self.update_output_text(system.get("profiler").output)
+        self.analysis_available = bool(system.get("profiler").output)
 
     def update_field_text(self, profilerDict):
         for k, v in profilerDict.items():
@@ -167,5 +168,4 @@ class ProfilerPage(UIConnectedWidget, UIPage, Events):
         self._fapd_profiler.stop_prof_session()
         self.running = False
         self.display_log_output()
-        self.analysis_available = True
         self.refresh_toolbar()
--- a/fapolicy_analyzer/ui/profiler_page.py
+++ b/fapolicy_analyzer/ui/profiler_page.py
@@ -14,14 +14,14 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import logging
-from events import Events
 from time import sleep
 
 import gi
+from events import Events
 from fapolicy_analyzer.ui.actions import (
-    set_profiler_state,
-    set_profiler_output,
     clear_profiler_state,
+    set_profiler_output,
+    set_profiler_state,
 )
 from fapolicy_analyzer.ui.faprofiler import FaProfiler
 from fapolicy_analyzer.ui.store import dispatch, get_system_feature
@@ -34,7 +34,9 @@ from gi.repository import Gtk  # isort: skip
 
 class ProfilerPage(UIConnectedWidget, UIPage, Events):
     def __init__(self, fapd_manager):
-        UIConnectedWidget.__init__(self, get_system_feature(), on_next=self.on_next_system)
+        UIConnectedWidget.__init__(
+            self, get_system_feature(), on_next=self.on_next_system
+        )
         self.__events__ = [
             "analyze_button_pushed",
             "refresh_toolbar",
@@ -92,10 +94,11 @@ class ProfilerPage(UIConnectedWidget, UIPage, Events):
             self.update_field_text(system.get("profiler").entry)
 
         buff = self.get_object("profilerOutput").get_buffer()
-        text = buff.get_text(buff.get_start_iter(), buff. get_end_iter(), True)
+        text = buff.get_text(buff.get_start_iter(), buff.get_end_iter(), True)
         if not text == system.get("profiler").output:
             self.update_output_text(system.get("profiler").output)
-        self.analysis_available = bool(system.get("profiler").output)
+            self.analysis_available = bool(system.get("profiler").output)
+            self.refresh_toolbar()
 
     def update_field_text(self, profilerDict):
         for k, v in profilerDict.items():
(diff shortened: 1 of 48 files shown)


dorschs57 (commented):
I don't think `MainWindow` is the right place to save the profiler state. If there is some state that needs to be persisted across the application, putting it in the application state via redux the the right place for it. I also feel like its not very intuitive to have to hit the "refresh" button to reload the profile data when I return to the screen. Maybe @jw3 or @tparchambault have a different opinion on this, but I envision the users interaction to be something like this:

1. Run some profile session
2. Go to Analyzer and do some analysis and make corrections
3. Go back to profile and previous session is already loaded and ready to go
4. Run profile session again
5. Go to Analyzer...
6. repeat, repeat, repeat

With saving the profiler data in the redux state you would also save off the previous output and reload that so when the user returns to the profiler it's as they left it when they went to the Analyzer. You could keep the "refresh" button so they can reload things if they change some settings and decide to go back to the previous state, but also add a "new" or "clear" button that would clear out the profiler completely and set it for a new run.

jw3:
> the users interaction to be something like this:
> 
>     1. Run some profile session
> 
>     2. Go to Analyzer and do some analysis and make corrections
> 
>     3. Go back to profile and previous session is already loaded and ready to go
> 
>     4. Run profile session again
> 
>     5. Go to Analyzer...
> 
>     6. repeat, repeat, repeat

concur

jw3 (approved):
Looks good :+1:

jw3:
See #635 for the next iteration

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -73,9 +85,26 @@ def __init__(self, fapd_manager):
...
         return self.analysis_available
 
+    def on_next_system(self, system):
+        profilerDict = system["profiler"]
+        self.update_field_text(profilerDict)
+
+    def update_field_text(self, profilerDict):
+        prefixes = ["execute", "arg", "user", "dir", "env"]
+        if not profilerDict:
+            profilerDict = {prefix + "Text": "" for prefix in prefixes}
```
dorschs57, line 95: If you set the default value in the `profiler_reducer` to the dict you need as a default you need, you wouldn't need to build one there. On line 35 of `profiler_reducer.py` change `{}` to `{<all the keys you need with values of empty string>}`.

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -73,9 +85,26 @@ def __init__(self, fapd_manager):
...
+        for prefix in prefixes:
+            buff = self.get_object(prefix + "Entry").get_buffer()
+            buff.set_text(profilerDict[prefix + "Text"],
+                          len(profilerDict[prefix + "Text"])
+                          )
```
dorschs57, line 100: If you changed the widget names in the glade to mirror the keys of the `profilerDict` you'd be able to do something like

```python
for k,v in profilterDict:
    self.get_object(k).get_buffer().set_text(v, len(v))
```
```

File: fapolicy_analyzer/ui/profiler_page.py [hunk 3]
```diff
@@ -73,9 +85,26 @@ def __init__(self, fapd_manager):
     def analyze_button_sensitivity(self):
         return self.analysis_available
 
+    def on_next_system(self, system):
+        profilerDict = system["profiler"]
+        self.update_field_text(profilerDict)
```
dorschs57, line 90: I'd suggest you keep `profilerDict` at the class level and then only call `update_field_text` if that dict is different than what you getting back from the `system` object here. It doesn't seem to be causing any issues, but the code the way you have it will do an extra call to `update_field_text` when you `dispatch(set_profiler_state(entryDict))` below.

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -90,7 +119,7 @@ def get_entry_dict(self):
             "dirText": self.get_object("dirEntry").get_text(),
             "envText": self.get_object("envEntry").get_text(),
         }
-
+        dispatch(set_profiler_state(entryDict))
```
dorschs57, line 122: If you implement the class level `profilerDict ` like suggested above you can set its value here and then dispatch it. That way in your `on_next_system` code the class object and the one coming in on system should be equal and you it will eliminate the extra refresh of the form data.

dorschs57 (changes requested):
Did you want to keep the results from the last profile run in the state object? That way when the user comes back to the profiler page it could be reloaded exactly as they left it. From there they could toggle back and forth to review the analyzer and the profiler output.

jw3:
> Did you want to keep the results from the last profile run in the state object?

That would be nice.  I was only reviewing the input params, but yeah, this sounds even better.

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -63,35 +68,64 @@ def __init__(self, fapd_manager):
...
         return self.analysis_available
 
+    def on_next_system(self, system):
+        if not self.inputDict == system.get("profiler").entry:
+            self.update_field_text(system.get("profiler").entry)
+
+        buff = self.get_object("profilerOutput").get_buffer()
+        text = buff.get_text(buff.get_start_iter(), buff. get_end_iter(), True)
+        if not text == system.get("profiler").output:
+            self.update_output_text(system.get("profiler").output)
```
dorschs57, line 97: If you did something like
```
self.analysis_available=bool(system.get("profiler").output)
```
here would it enable the Analyzer button when there was existing output? That way if the user came back here to look at some output they could go back to the Analyzer using the toolbar button.

dorschs57 (approved):
All the changes look good and I think provide a pretty good user experience. I had one question/possible tweak. If it works out I think that would be useful, otherwise I'm good with this one.

jw3:
@egbicker I'll wait for your response or commits for the last comment.  Let me know when it's good and I'll merge and pick into release.

egbicker:
@jw3 Just gave @dorschs57 request a spin and it looks good. Once this CI finishes it should be good to go

fapolicy_analyzer/ui/profiler_page.py
dorschs57: This doesn't seem to be working. Do we need a toolbar refresh to get the button to become active? I
1. run profiler
2. clicked to go to Analyzer
3. went back to the profiler
4. the entries and output were reloaded, but the Analyzer button was not active until I ran again

dorschs57: Also the persistent checkbox does not seem to get repopulated.

dorschs57:
@egbicker give that last update I pushed a spin and see if its good. It seems to work for me.

fapolicy_analyzer/ui/profiler_page.py
egbicker: I'm seeing the markup show as plaintext instead of the bolding

dorschs57: > I'm seeing the markup show as plaintext instead of the bolding

That should be fixed with my last commit.

dorschs57:
@jw3 I think this is ready for merging and 🍒-picking now.

jw3 (approved):
Looks good! :trophy:

PR #630, opened 2022-08-26
Disable Profiler menu item for non-root users

tparchambault:
Add development overrides via `PROF_UI_ENABLE` env var and/or the existence of the magic file, `/tmp/prof_ui_enable`.


tparchambault:
Overrides to enable the Tools > Profiler menu-item during development:

1. Set the environment variable to anything except FALSE (case insensitive).
2. Create the file `/tmp/prof_ui_enable`

Unit-tests are still on the agenda.

dorschs57 (approved):
Test fine on my dev env.

PR #631, opened 2022-08-26
Adding analyzer button sensitivity

egbicker:


tparchambault:
Sensitivity appears to be correct, however the application is not automatically transitioning to the Analysis tool pane.

Tested over FC34 via an rpm install.

tparchambault:
All good over FC34 via rpm install as the root user.

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -32,6 +32,7 @@ def __init__(self, fapd_manager):
         UIConnectedWidget.__init__(self, get_system_feature())
         self.__events__ = [
             "analyze_button_pushed",
+            "refresh_toolbar",
```
tparchambault, line 35: Maybe at some point, "refresh_toolbar" should be converted to something on the order of an event, e.g. "analysis complete" which might map into a redux state var if needed.

tparchambault (approved):

jw3 (approved):
Functions as expected

jw3:
Found #632 while testing this.

jw3:
@dorschs57 are you good with this?

dorschs57 (approved):
I'm OK with this. I'm not a big fan of needing to reach back to the `MainWindow` to force a toolbar refresh, but I can't think of a better way to do it right now. Ultimately I think the toolbar and menu items need to be wired into redux like we talked about before, but that's a bigger issue.

egbicker:
> I'm OK with this. I'm not a big fan of needing to reach back to the `MainWindow` to force a toolbar refresh, but I can't think of a better way to do it right now. Ultimately I think the toolbar and menu items need to be wired into redux like we talked about before, but that's a bigger issue.

Completely agree

PR #633, opened 2022-08-31
Exclude tests from wheel

### LINKED ISSUE #618: RPM contains test by-products
image

jw3:
Explicitly include the packages we want instead of globbing.

Closes #618

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ setup(
             "fapolicy_analyzer.css",
             "fapolicy_analyzer.glade",
             "fapolicy_analyzer.redux*",
+            "fapolicy_analyzer.resources*",
             "fapolicy_analyzer.ui*",
             "fapolicy_analyzer.util"
         ],
--- a/setup.py
+++ b/setup.py
@@ -40,12 +40,8 @@ setup(
     packages=find_namespace_packages(
         include=[
             "fapolicy_analyzer",
-            "fapolicy_analyzer.css",
-            "fapolicy_analyzer.glade",
-            "fapolicy_analyzer.locale",
-            "fapolicy_analyzer.redux",
-            "fapolicy_analyzer.resources",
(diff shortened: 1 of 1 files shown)


jw3:
@dorschs57 do you see any issues with the changes

@tparchambault please check the rpm install for function

setup.py
```diff
@@ -38,8 +38,15 @@ def get_version():
     name="fapolicy-analyzer",
     version=get_version(),
     packages=find_namespace_packages(
-        include=["fapolicy_analyzer", "fapolicy_analyzer.*"],
-        exclude=["*.tests"],
+        include=[
+            "fapolicy_analyzer",
+            "fapolicy_analyzer.css",
+            "fapolicy_analyzer.glade",
```
dorschs57, line 44: I think you need to include "fapolicy_analyzer.locale*" in here.

jw3: There is some weirdness between how `package_data` is mapped and the package search.  I'll add it for clarity, but it appears to be included from the `package_data` entry alone.

Noting for posterity; I saw that the other resources, which are also sorta "package data", can't be excluded from the  `packages` entry.  Is that perhaps because they are "remapped" by globbing from their actual locations to packages under `fapolicy_analyzer`, while the locale is not (it uses the absolute path rather than a glob)?  Idk.

As long as it works I'm not concerned right now, though someday we should survey other projects to determine if there is some tweaks we could make to these mappings for correctness / clarity.

jw3:
@tparchambault this is ready to give it a sanity check and make sure everything still functions.

tparchambault (approved):
Tested via rpm installs over FC34 and RHEL8.6. No issues observed.

Independent of this ticket, verified that the `epel-release`pkg must be manually installed on RHEL8 prior to the rpm installation of the fapolicy-analyzer.

PR #580, opened 2022-08-03
Build dependency updates

jw3:
We have some gaps in build dependencies that block going to EPEL.

# See the WIP example solution here

- https://github.com/jw3/example-python-rust-copr
- https://copr.fedorainfracloud.org/coprs/jwass3/example-rust-python/

# Try the WIP package here

```text
dnf copr enable jwass3/example-rust-python 
dnf install python3-rulec
python3 -m rulec /etc/fapolicyd/rules.d
```

# Build Time

### Package misses: 18 -> 7
```
	pyflakes               -> 2.4.0
	typed-ast              -> 1.5.4
	pyparsing              -> 3.0.9
	argparse               -> configargparse 1.5.3
	semantic-version       -> 2.8.4
	pycairo                -> 1.21.0
	callee                         X
	autoflake                      X
	flake8-quotes                  X
	pytz                   -> 2022.1
	babelgladeextractor            X
	babel                          X
	pytest                 -> 7.1.2
	dataclasses                    X
	pygobject              -> 3.42.2
	flake8-i18n                    X
	mypy-extensions        -> 0.4.3
```

### Major version misses: 5
```
	attrs:21               -> 22.1.0
	flake8:4               -> 4.0.1
	importlib-metadata:0   -> 4.12.0
	tomli:1            -   -> 2.0.1
```

### Package misses: 8
```
	hermit-abi                X
	indoc-impl                X
	lmdb-rkv-sys              X
	lmdb-rkv                  X
	ariadne                   X
	redox_syscall             X
	redox_users               X
	paste-impl                X
```
### Major version misses: 5
```
	nom:6               -> 7.1.1
	paste:0             -> 1.0.7
	os_str_bytes:3      -> 6.2.0
	directories:2       -> 4.0.1
	indoc:0             -> 1.0.7
```

PR #649, opened 2022-09-10
Build SRPM

### LINKED ISSUE #506: Modify jinja template spec file to have target OS specific dependency requirement lists.
Fedora34 and RHEL8 have differing versions of python and slightly different package requirements. FC has some python features installed as part of the base package, while RHEL relies on some backported packages contained in EPEL.

Also dependent runtime packages should be required and not recommended. These are hard dependencies, i.e. the application terminates with an exception when the import fails.

### LINKED ISSUE #622: Update CI targets
- Add Rawhide as a build target
- Add 36, 37, drop 34, 35

### LINKED ISSUE #627: Add Fedora Copr build
Add Fedora Copr build for RPM builds and non-official distribution method.

The goal is to achieve an offline RPM build that prepares us for vendoring artifacts in a `fedpkg`/`rhpkg` builds.

A side effect of having this Copr build is that we end up with a `dnf` only method of distribution

```
$ dnf copr enable ctc-oss/fapolicy-analyzer 
$ dnf install fapolicy-analyzer
$ fapolicy-analyzer
```

See an example here: https://github.com/jw3/example-python-rust-copr

See #580

### LINKED ISSUE #637: Add branch pattern to produce rpms
Add a branch pattern to the RPM CI to produce builds without the need for a PR.

### LINKED ISSUE #641: Playbook does not reload daemon
The fapolicyd play should ensure the daemon has been reloaded and then has accepted all changes made.

Currently there are changes that occur after the reload, setting the daemon up for failure at first next start/stop operation invoked by the analyzer.

jw3:
The goal of this PR is to resolve the build-time dependency issues encountered with both Rust crates and Python packages.

To address the dependency issues we have implemented a vendoring process for Rust that vendors the crates that have no official packages and lists the available crates as `BuildRequires`.  On the Python side we have several items missing in RHEL builds which we address by listing SourceX dependencies from PyPI and installing those in the `%prep` stage.

This process is tested end-to-end by generating a SRPM which is then used to create a RPM in an offline build.  The GitHub CI process implements this and there is also a Fedor Copr build that also proves out the approach.

The main product of this effort is the **spec file**, located at [**scripts/srpm/fapolicy-analyzer.spec**](https://github.com/ctc-oss/fapolicy-analyzer/blob/build/srpm/scripts/srpm/fapolicy-analyzer.spec).  The missing pieces that we vendor are what we are referencing when saying _there is a gap in the build-time dependencies_.

This PR resolves the build time dependency issues, and a few other spec and rpm build related issues along the way.  The process defined here accomplishes that by vendoring the Python and Rust sources that are without official support into the SRPM.  The SRPM is then be used to create both Fedora and RHEL 8 RPMs which use official packages where possible and vendored packaged where not.

**There are no run-time dependencies being vendored**.

The original RPM build and CI is left undisturbed by this PR, in order to preserve for comparison.  It will likely be deprecated after this one is merged.

Much of this was prototyped at https://github.com/jw3/python-rulec and some in #580

## todo

- [x] add vendor prefix to vendored tarballs
- [x] add discriminator between the all crates and partial crates tarball
- [x] exclude the debug* rpms from the artifacts
- [x] create CTC Copr account 
- [x] update playbook to use CTC Copr rather than jwass3
- [x] document the remaining vendored python pieces
- [x] Update wiki with Copr info
- [x] Tie Copr Makefile to the root Makefile using common targets (eg. dist) (defer to separate issue)
    - #652 
- [x] Split RHEL from FC in the Spec to reduce review complexity
- [x] Add comment on the vendor tarball sourcex entry to describe what how why
- [x] Remove travis.yml
- [x] Remove original RPM build and scripts
- [x] Remove Bats tests
- [x] Update wiki to reflect removal of Bats tests
- [x] Update wiki to reflect changes to RPM building

## Rationale

The following documents the what and why of this PR to make reviewing it easier.  There is a lot of ground covered in this one, and some of the process may still not be the right way, but I want to highlight why it is done the way it is done, in case this way is the only way.

### The What
1. Adds a spec file that supports Fedora and EL which supports offline builds using vendored Python and Rust
1. Add scripts to vendor all or a subset of the Rust crate dependency, the remainder are added as SourceX
2. Add scripts to determine what crates need vendored based on the availability of official packages
3. Add SourceX dependencies for Python build tools unavailable on RHEL
4. Add Makefile to configure Fedora Copr build environment
5. Add GitHub CI workflow to build SRPM, and then rebuild a RPM from it
6. Add matrix to the SRPM workflow to run on fc:37, rawhide, and UBI 8
7. Capture SRPM build artifacts in the releases, including spec file, srpm, rpm, vendor tarballs
8. Rewrite the spec file captured in release to point to the url of the release artifact rather than the local path

### The Why

1. There are build time dependencies missing for Rust in both Fedora and RHEL
2. There are build time dependencies missing for Python in RHEL
3. We must be able to build offline, using vendored dependencies, from a SRPM
4. We need to build for RHEL 8 for distribution, but also want to prepare for next-gen features from Fedora Rawhide

### Other Whats
1. Container file SRPM builds for RHEL and Rawhide
2. Vagrant Rawhide configuration
3. Playbook will provision from Copr when version is `latest` (default if unspecified)
4. Playbook will provision from GitHub when version is `latest-release` or a specific release (eg. `v0.6.0`)

### What about using [rust2rpm](https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/)?

It appears the primary focus of the `rust2rpm` tools is to produce RPMs for Rust devel crates.  It's not that it blocks the ability to produce Rust applications, but it makes some assumptions that produce configuration from the macros that cause issues in our case.  We are a Rust app, not a crate. 

One example is `%cargo_generate_buildrequires` is very convenient, it however injects crate dependencies on all crates rather than all non-vendored crates.

Another is `%crates_source` macro is useful to download crates as source entries.  This was used in an early version, however there are a couple of problems.  

1. You must still extract the crate into the registry for it to be used
2. The crate extract will be missing a required checksum file which you must create (though it can be empty)

Given the additional noise these two items create it ends up just as easy to vendor a bundle crates in as source, which also gives us the ability to use `cargo-vendor-filterer` to reduce dependencies that are not needed for our target.

We do use the `%cargo_prep` macro.  It is useful to init the .cargo config 99% of the way, and a quick `sed` call can then  patch the macro generated config.

### Why is the SourceX in the released artifacts is different than the master branch SourceX?

The spec in `scripts/srpm` has Source{0,1} set to `fapolicy-analyzer.tar.gz` and `crates.tar.gz` respectively.  This supports two use cases (1) SRPM packaging (2) use in development environments.

The third use case to support is building the RPM using HTTPS sources.  We support that for published releases only.  The SRPM CI vendors the deps and builds the spec, producing the artifacts.  Then the CI  replaces the SourceX in the spec with the published locations of the release artifacts.  This results in a fully qualified spec, relative to the release it is built in, that gets published as part of a release and that lives beside the published vendored artifacts.  The intent here is to make RPM builds where internet is available easy with no need to provide any vendored sources yourself.

### Where can the SRPM be found?

Given a GitHub release of fapolicy-analyzer you can either download the SRPM directly from the release artifacts, or download the spec and pull the associated SourceX which are linked to release artifacts. 

### Why vendor Rust crates?

1. We must produce offline capable SRPM builds.
2. We have crate dependencies that do not have official packages.
3. We do not (at this time) intend to submit the crates from (2) to have packages created.
4. The alternative here is to vendor the missing dependencies (in the case of RHEL that is all of them).

### Why vendor Python packages?

There is one Python build-time dependency that we require, `setuptools-rust`, that is not available as a rpm package.

There are then a few additional dependencies due to version numbers required to build using `setuptools-rust`, so we end up with

1. `setuptools-rust`: unavailable on RHEL - we rely on it to build the PyO3 bindings.
2. `setuptools`: the packaged version 39.x does not meet requirements of `setuptools-rust`
3. `semanticversion`: unavailable on RHEL, required by `setuptools-rust`
4. `pip`: the packaged version at 9.x will not install `setuptools-rust`
5. `wheel`: required to upgrade `setuptools`
6. `setuptools_scm` 6.4.2
7. `semantic_version` 2.8.2
8. `packaging` 21.3
9. `pyparsing` 2.1.0
10. `tomli` 1.2.3
11. `flit_core` 3.7.1
12. `typing_extensions` 3.7.4.3
13. `pytz` 2017.2
14. `Babel` (linked from dnf install)

Adding these to the build was tedious, but removing the `setuptools-rust` usage would be more so.

### Why bundle the vendored crates rather than leaving them as individual `*.crate` tarballs?

The examples we have observed in the wild, for example with Rust, bundle all crates together using `cargo vendor`.

An early version of the PR listed each missing crate as SourceX using `%crate_source` from rust2rpm, but that has some drawbacks

1. Additional noise from manually unpacking each crate into the source dir.
2. There is a checksum file that must be touched in each extracted crate, even more noise.
3. We can use cargo tooling to filter down to our platform specific crates, using `cargo-vendor-filterer`

### What are the build-time differences between Fedora and RHEL?

Rawhide has many Rust capabilities, including a good selection of popular crates and also on the Python side there is a package for `setuptools-rust`, which is big for us and prevents the cascading of additional dependency vendoring and the increased complexity in the build which RHEL experiences.

So the first major difference is that for Fedora we do not vendor any Python.

The second is that we do not vendor all Rust crates, the majority of them are available through official packages.  We vendor the remainder and mix them in.

Significant, but manageable.

### Why are there two specs?

The "production" spec now lives under the project root, this is currently FC only.  The scripts/srpm spec is the next version with RHEL support.  Keeping both to ensure we don't have to deal with any drift in the future when RHEL moves into the production spec.

### 

## Testing

To install from Copr on RHEL 8 or FC >= 37

```text
dnf install -y dnf-plugins-core
dnf copr enable -y ctc-oss/fapolicy-analyzer
dnf install -y fapolicy-analyzer
```

The Vagrant provisioning will also install from Copr if no version is specified.

or from Podman

```text
make -f .copr/Makefile vendor
podman build  -t fapolicy-analyzer:el -f scripts/srpm/Containerfile.el .
podman run --rm  -it --network=none fapolicy-analyzer:el
```

---

Closes #506
Closes #622
Closes #627
Closes #637 
Closes #641


jw3:
image

03db61e9

jw3:
@tparchambault try the rhel and fedora playbooks on this branch to test the application function.  See if you notice anything end to end that is off.

If you do not specify a version when provisioning the playbook will pull latest from Copr.

jw3:
This is also removing some deprecated components that were flagged in Coverity scans.

PR #650, opened 2022-09-12
Multiple gid in analysis

### LINKED ISSUE #267: Support multi group bindings
In #261 support was added for group lists in event logs.  The bindings and python need updated to handle this list, where now they are just using the first item in the list.

jw3:
Ensure gid lists in analysis logs make it to the UI.

## Testing

Given the log

```
rule=1 dec=allow perm=any uid=0 gid=990,999 pid=111 exe=/usr/bin/bash : path=/usr/bin/foo
rule=1 dec=allow perm=any uid=0 gid=999     pid=111 exe=/usr/bin/bash : path=/usr/bin/bar
```

```
Subjects in log: 1

# Subject events

# User events - 0
0:990 /usr/bin/bash => /usr/bin/foo
0:999 /usr/bin/bash => /usr/bin/foo
0:999 /usr/bin/bash => /usr/bin/bar

# Group events - 999
0:999 /usr/bin/bash => /usr/bin/foo
0:999 /usr/bin/bash => /usr/bin/bar

# Group events - 990
0:990 /usr/bin/bash => /usr/bin/foo
```

`python3 examples/analyze_log.py tests/data/events2.log`

Closes #267

### DIFF
--- a/crates/pyo3/src/analysis.rs
+++ b/crates/pyo3/src/analysis.rs
@@ -47,7 +47,7 @@ pub(crate) fn expand_on_gid(rs: &Analysis) -> Vec<PyEvent> {
             },
         })
     }
-    dbg!(&r);
+    //dbg!(&r);
     r
 }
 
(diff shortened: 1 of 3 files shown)


jw3:
@tparchambault @dorschs57 could you take a quick look at this, check that it doesn't break or introduce odd behavior to the analysis view.

PR #651, opened 2022-09-22
Include User Guide under Help menu

### LINKED ISSUE #645: Help manual needs to be included under Help menu
We need an offline copy of the user manual.

1. Maintain the docs as the GitHub Wiki
2. During RPM CI builds we clone the Wiki and build the User Guide section Markdown into HTML
3. Bundle the HTML as the doc
4. Have a placeholder HTML for non-RPM (like dev env) that just has a link to the Wiki

dorschs57:
Adds a user guide entry under the help menu.

- Generates the user guide from the online wiki documentation
- Integrates the user guide build process into CI and the RPM build
- Installs the user guide to /usr/share/help via RPM
- Internationalization support is included for the documentation
- Adds target to root Makefile that pulls and builds help locally
  - `make help-docs`
  - Useful for contributing translations

Closes #645


fapolicy_analyzer/setup_cmds.py
```diff
@@ -0,0 +1,208 @@
...
+from distutils import dir_util
+from distutils.cmd import Command
+from glob import glob
+from os import path
+from pathlib import Path
+
+
+class parse_help(Command):
+    description = "Update source controlled help documentation files from online source"
+    user_options = [
```
jw3, line 28: We will need some way to filter wiki paths... including/excluding individual paths may become tedious and error prone.

I have a few thoughts, nothing that seems foolproof.

The easiest one is adding a hash tag identifier to the pages that need included.

Eg.
```
My Wiki Page
===

- one thing
- two thing
- three thing

#user-guide
```

dorschs57:
@jw3 here are some links to what I was talking about with how meld is handling putting help docs out as part of the `data_files` argument to `setup`.  

https://github.com/GNOME/meld/blob/3ee63e9f2f88dfe3acee3c5582c59e107b755997/setup.py#L70
https://github.com/GNOME/meld/blob/3ee63e9f2f88dfe3acee3c5582c59e107b755997/meld/build_helpers.py#L229

It appears the file path they provide for the help data files is `share/help/C/meld` where the `C` directory is the location for the base (non-translated) docs. When I look at the location from my rpm install though they end up in `/usr/share/help/C/meld`. Not sure how that gets worked out.

jw3:
Should be good with a cmdclass like this in dev, but the rpm build will do something a bit different.

dorschs57:
This will require some more research. I just added a custom filter so any url pointing to one of our wiki pages is changes to a relative path to the converted docbook page. Now that I can navigate around a bit, there is an issue when you go from one docbook file to the next. There is no automatic link back to the previous page. I think what we probably need to look into is converting all our markdown files into a single docbook file. I guess our other options is to step back from docbook format and see if we can get HTML working within yelp.

scripts/srpm/fapolicy-analyzer.spec
```diff
@@ -229,6 +247,7 @@ alias python3=%{venv_py3}
 %endif
 
 python3 setup.py compile_catalog -f
+python3 setup.py build_help
```
jw3, line 250: This cannot run here since we cannot rely on a network connection during hte RPM build.

Instead the compiled documentation archive should be added as a SourceX file and unpacked to the correct filesystem location during the RPM build.

fapolicy_analyzer/setup_cmds.py
jw3: To simplify this we are going to do a single page user doc on the wiki.  Future use may require a more complicated layout, but this should suffice for the foreseeable future.

dorschs57:
@jw3 I think I have this to the point where the artifact from the `help_build` CI job needs to be incorporated into the spec file. Can you take a look at that part of it? What is packaged in the artifact needs to be deployed to `/usr/share/help`. The `C` directory to `/usr/share/help/C/fapolicy-analyzer` and the `es` (or any other language) to `/usr/share/help/es/fapolicy-anayzer`.

dorschs57:
@jw3 I looked into why `markdown2` and `beautifulsoup4` are required as build dependencies during the rpm build. It's because the .spec file calls `python3 setup.py compile_catalog -f` during the `Rebuild SRPM` step. To load `setup.py` we'll need those python dependencies. I have them working in the FC37 and FC38 builds, but RHEL8 is still failing. Maybe you can take a look at that?

dorschs57:
I put a `.env` file under the help directory that will hold the `HELP_COMMIT` env. This can be updated whenever we get a new user guide commit in the wiki. Using pipenv automatically loads this file and exports the envs.

I also update the build help CI step to tar the files and then output that as an archive. This will still need worked into the RPM build process.

setup.py
```diff
@@ -69,4 +71,5 @@ def get_version():
         "fapolicy_analyzer.resources.sourceview.language-specs": ["*.lang"],
         "fapolicy_analyzer.resources.sourceview.styles": ["*.xml"],
     },
+    cmdclass={"build_help": build_help},
```
jw3, line 74: What was the advantage of making build_help a setup cmd vs. just a plain old script?

dorschs57: Well the original thought way back when was this would run as part of the of RPM build like the `compile_catalog` command. The nice thing about the current setup is a developer doing translations or running locally just needs to run `python setup.py build_help` and they will have a copy of the help locally. Are you thinking we make this a standalone script?

jw3: > @jw3 can you handle updating the RPM build process?

Yeah looking now.

Can you show a quick example of

1. Call the script that clones and downloads the images
2. Call the `build_help` using the files from (1)

dorschs57: Sure. 
1. To download a local copy of the help file and convert it to HTML call `pipenv run python setup.py update_help`. The defaults will place it under the project root in the `help` directory. This will also create the portable translation file that can be used for doing translations.
    - If you need to pull the help for a specific commit update the commit in the `help/.env` file and export `export PIPENV_DOTENV_LOCATION=help/.env ` or use the `-c` flag when calling the command giving the commit number.
    - If you need to use a proxy server for downloading use the `-p` flag when calling the command giving the URL of the proxy server.
2. To build the help for release, run `pipenv run python setup.py build_help`. By default this will use the `help` directory as source and output to `build/help`. This also performs any translations and places the translated files in the output.

jw3: Thanks.  Sounds good.  The path forward is

1. Run step 1 outside the RPM build, tar up the results, pass in as SourceX
2. Run step 2 in the RPM build and install the results

I'd forgot about the .mo previously.  We cant have binaries in the srpm.

I'll update the RPM build.

jw3:
image

jw3:
Pretty close.  Can build rpm in and run from container. 

RHEL CI is blocked due to the setup.py requiring setuptools-rust, which on RHEL is not available.  FC is blocked for related reason, though it could be unblocked by installing it.

The best solution build-wise is to eliminate the dependency on setup.py to build the doc.  I will comment above in the thread where we were talking about example help commands.

setup.py
jw3: @dorschs57  Can you give me a quick example of doing `update_help` without going through setup.py?

Earlier I overlooked an important point.. that we also need to decouple the help doc download from the setuptools-rust dependency (which is not available on RHEL).  As long as the download call is made from setup.py this cant happen.

fapolicy_analyzer/glade/main_window.glade
```diff
@@ -229,20 +235,29 @@
...
                     <child type="submenu">
                       <object class="GtkMenu">
                         <property name="visible">True</property>
-                        <property name="can-focus">False</property>
+                        <property name="can_focus">False</property>
+                        <child>
+                          <object class="GtkMenuItem" id="helpMenu">
+                            <property name="visible">True</property>
+                            <property name="can_focus">False</property>
+                            <property name="label" translatable="yes">Documentation</property>
```
jw3, line 249: How about "User Guide"

setup.py
jw3: @dorschs57 

I tried several ways to run the update outside of setup.py, got the clone but not the conversion.  That logic down in the command class couldnt be hijacked with the time available right now.

My preference is to convert the commands into standalone scripts and call them from the Makefile rather than setup.py.  If that is possible I think we can still give the user a simple way to build all this stuff, eg `make docs` and avoid the issues with setup.py dependencies.

fapolicy_analyzer/glade/main_window.glade
dorschs57: Done.

setup.py
dorschs57: @jw3  I updated the setup cmds to a standalone script called `help_utils.py`. You should now be able to run `python help_utils.py update` to update and `python help_utils.py build` to build and I think the defaults will work. Again the commit can be changed by updating the hash in the `help/.env` file.

jw3:
Working. Noteworthy items are:
- removed dotenv, there was a dependency conflict on rhel that i didnt sort out, just dropped it and put ref in a file
- moved some imports into the functions they are called from to prevent coupling of unrelated dependencies (if the single utils were split apart it would resolve this, not a blocker)
- moved the help_utils script into an __init__ and __main__ under help so you can call it like `python3 help build`
- moved the github wf over to rocky due to some package issues
- cleaned up a few sharp corners in the py and rs vendoring process 

Also for posterity:  To render the help in webkit we end up depending on a bunch of other libraries, ~80/260 MB worth.  But yelp wont help this out any, it depends on the same webkit2gtk3 dependency and all the transients from it.  So basically a wash.  To render html help with Gtk this is the price we will pay.  Is there another way to render help?  Doesnt need answered now, but perhaps someday it will.

jw3:
@dorschs57 do you have anything left on your todo list for this one?  I'll merge it later today if you are good with it.

dorschs57:
@jw3 I'm good here. Merge when ready.

scripts/srpm/fapolicy-analyzer.spec
```diff
@@ -10,21 +10,24 @@ Source0:       fapolicy-analyzer.tar.gz
...
-Source5:       %{pypi_source wheel 0.37.0}
-Source6:       %{pypi_source setuptools_scm 6.4.2}
-Source7:       %{pypi_source semantic_version 2.8.2}
-Source8:       %{pypi_source packaging 21.3}
-Source9:       %{pypi_source pyparsing 2.1.0}
-Source10:      %{pypi_source tomli 1.2.3}
-Source11:      %{pypi_source flit_core 3.7.1}
-Source12:      %{pypi_source typing_extensions 3.7.4.3}
-Source13:      https://files.pythonhosted.org/packages/source/p/pytz/pytz-2017.2.zip
+Source10:       %{pypi_source setuptools-rust 1.1.2}
```
jw3, line 19: Bumped these up by several numbers to avoid bumping them all again every time a lower numbered source is added.  Might never happen again, but it could.

dorschs57:
@jw3 I made the changes to display a static file with a link to the online users guide if the local help files are not available. The RPM build is broken this morning though. It looks like the Rocky 8.6 appstream repo you just updated to is now gone as of this morning. We'll either need to update to Rocky 8.7 or point to the archive repo at https://dl.rockylinux.org/vault/rocky/8.6/AppStream/.

jw3:
The explicit add of the appstream can go away since the build switched from ubi to rocky. That it remained was an oversight.  It was a convenient time for the failure :boom: :trophy:

PR #656, opened 2022-10-14
Fix profiler path validation

### LINKED ISSUE #655: Profiler executable argument validaiton should consider PATH env
The profiler will not allow non-absolute paths to be used as the executable.  The expectation is that the PATH in the environment (either specified or inherited) would be used when looking for the executable.

For example use `ls` as the target will fail,  you instead must use absolute path of  `/bin/ls`.
The validation should consider whatever path the execution is going to use when validating the path variable.

This is a regression from #640 

Blocking release of v0.6.1

tparchambault:
Support for validating relative executable paths by using the existing PATH from the users environment.  This also supports evaluating a custom PATH provided through the profiler GUI.

This fixes the issue from #655 where users are forced to use absolute paths to avoid a validation error reporting that the file cannot be found.

Closes #655


tparchambault:
Target executables specified with relative paths are now supported, whether the PATH environment variable is acquired from the pkexec runtime environment (On RHEL8.6: `/usr/sbin:/usr/bin:/sbin:/bin:/root/bin`) or provided by the user in the enironment variable entry in the Profiler Page UI.

If the PATH env var provided by the user on the Profiler Page, the below syntax is supported and expanded in the following order::

- `$PATH` or `${PATH}` are expanded to the pkexec runtime environment's PATH value.
- Single periods, lone leading colons, lone trailing colons, and internal double colons in the `PATH` variable are expanded to the user-specified Profiler Page `Working Directory` entry. If not specified uses the directory where fapolicy-analyzer was invoked.
- Double periods are expanded to the parent of the user-specified Profiler Page `Working Directory` entry otherwise the parent of the directory where invoked.

tparchambault:
If you aren't running as `root` you can activate the Profiler Page menu item by creating a file, `/tmp/prof_ui_enable.` Alternatively, you can extract the attached tarball within your `/tmp` directory as it contains the above mentioned file and a trivial shell script, `Now.sh` the prints out a teletype test, and the environment variables `WEINER` and `PATH.`

Also, if not running as `root` the `fapolicyd` daemon can not be started/stopped by the `fapolicy-analyzer`, so only the execution of the specified target will occur.

[FaProfilerTestCode.zip](https://github.com/ctc-oss/fapolicy-analyzer/files/10035900/FaProfilerTestCode.zip)

fapolicy_analyzer/ui/faprofiler.py
```diff
@@ -56,51 +110,69 @@ def __init__(self, msg="Unknown error",
...
-        self.listCmdLine = [self.execPath] + self.execArgs.split()
+        self.timeStamp = None
 
         # euid, working directory and environment variables
         self.user = dictProfTgt["userText"]
-        self.pwd = dictProfTgt["dirText"]
+
+        # Set pwd - Use current dir for pwd if not supplied by user
+        tmp_d = dictProfTgt["dirText"]
+        self.pwd = tmp_d if tmp_d else os.getcwd()
```
dorschs57, line 123: If you are worried about this being empty this would be the safest and simplest way to do that. In fact all your initial accesses to the `dictProfTgt` dictionary should use the `.get()` method in case a bad dictionary is passed in that doesn't have all the keys needed.

```suggestion
        self.pwd = dictProfTgt.get("dirText") or os.getcwd()
```

fapolicy_analyzer/ui/faprofiler.py
```diff
@@ -37,17 +40,68 @@ class ProfSessionArgsStatus(Enum):
...
+        logging.debug("Internal double colon detected")
+        expanded_path = regexDoubleColon.sub(":.:", expanded_path)
+
+    # Expand periods with user supplied absolute path if provided otherwise
+    # use cwd from runtime environment
+    if cwd != ".":
+        expanded_path = re.sub(r":\.$", f":{cwd}", expanded_path)
+        expanded_path = re.sub(r"^\.:", f"{cwd}:", expanded_path)
+        expanded_path = re.sub(r":\.:", f":{cwd}:", expanded_path)
+        expanded_path = re.sub(r"^\.$", f"{cwd}", expanded_path)
```
dorschs57, line 94: I haven't tested this but I think you could replace this entire chunk with these lines below and end up with the same thing.

```suggestion
    expanded_path = re.sub(r":\.{0,1}$", f":{cwd}", expanded_path)
    expanded_path = re.sub(r"^\.{0,1}:", f"{cwd}:, expanded_path)
    expanded_path = re.sub(r":\.{0,1}:", "f":{cwd}:"", expanded_path)
    expanded_path = re.sub(r"^\.$", f"{cwd}", expanded_path)
```

fapolicy_analyzer/ui/strings.py
```diff
@@ -99,7 +99,8 @@
 PROF_ARG_OK = _("Profiler Session arguments are valid.")
 PROF_ARG_EXEC_EMPTY = _("Executable field is empty")
 PROF_ARG_EXEC_DOESNT_EXIST = _(": file does not exist.")
-PROF_ARG_EXEC_NOT_EXECUTABLE = _(": file is not executable")
+PROF_ARG_EXEC_NOT_EXEC = _(": file is not executable")
+PROF_ARG_EXEC_NOT_FOUND = _(": command not found")
 PROF_ARG_USER_DOESNT_EXIST = _(": user does not exist")
 PROF_ARG_PWD_DOESNT_EXIST = _(": working directory does not exist.")
 PROF_ARG_PWD_ISNT_DIR = _(": working directory is not a directory.")
```
dorschs57, line 106: Bit of a nitpick here, but some of these error messages end in a period and some do not.

fapolicy_analyzer/tests/test_faprofiler.py
```diff
@@ -246,3 +294,32 @@ def test_validateArgs():
...
+        "executeText": "ls",
+        "argText": "",
+        "userText": os.getenv("USER"),
+        "dirText": os.getenv("HOME"),
+        "envText": "",
+    }
+
+    dict_valid_args_return = FaProfSession.validateArgs(dictArgs)
+    assert len(dict_valid_args_return) == 1
+    assert ProfSessionArgsStatus.OK in dict_valid_args_return
```
dorschs57, line 325: I'd like to see this as 2 different tests. You can use pytest parameterize capabilities for this like below. You should alsos add a set of parameters for this to test a `dict_args` with missing or invalid keys.

```suggestion
@pytest.mark.parametrize(
    "dict_args, expected_status",
    [
        (
             {
                "executeText": "Now.sh",
                "argText": "",
                "userText": os.getenv("USER"),
                "dirText": "/tmp",
                "envText": "PATH=.:${PATH}",
            },
            ProfSessionArgsStatus.EXEC_NOT_FOUND,
        ),
        (
            {
                "executeText": "ls",
                "argText": "",
                "userText": os.getenv("USER"),
                "dirText": os.getenv("HOME"),
                "envText": "",
            },
            ProfSessionArgsStatus.OK,
        ),
    ],
)
def test_validateArgs_relative_exec(dict_args, expected_status):
    dict_valid_args_return = FaProfSession.validateArgs(dict_args)
    assert len(dict_valid_args_return) == 1
    assert expected_status in dict_valid_args_return
```

dorschs57 (changes requested):

fapolicy_analyzer/ui/strings.py
tparchambault: Made consistent; removed periods.

fapolicy_analyzer/ui/faprofiler.py
tparchambault: I do this multi-line thing so that my single lines are fully visible on an 80 char screen i.e. avoiding

> `<tab><tab>...<tab>self.pwd = dictProfTgt["dirText"] if dictProfTgt["dirText"]  else os.getcwd()`

which overflows. Was not worried about missing K=V pairs; if it's missing I want it to crash. If any of the UI fields are empty, the assoiated dict value is an empty string.

In any event, your suggestion: `self.pwd = dictProfTgt.get("dirText") or os.getcwd()` is better. Changed it.

fapolicy_analyzer/ui/faprofiler.py
tparchambault: Seems to work fine. I'll test it more on an rpm install platform. Thank you.

jw3:
Is this ready?

tparchambault:
Opened #670 related to the Profiler's default `pwd` given that within the pkexec environment where `/root is the default `pwd.`

fapolicy_analyzer/tests/test_faprofiler.py
tparchambault: Used the pytest parameterize mechanism, and also added code to detect if received dict's keys match expected dict's keys.

dorschs57 (approved):

jw3:
@tparchambault conflicts in the main .pot

jw3 (approved):
Works as expected :+1:

jw3:
Resoled the conflict, was just the file info date/time conflicting.

Issue #465, opened 2022-04-05
Time range selection for policy analysis


jw3:
1.  The temporal filter that is applied to the analysis
2. GUI for configuring the temporal filter

From the prototype document 
![image](https://user-images.githubusercontent.com/1545372/161842402-f52e657c-edd8-4bfc-9ac7-3cfdcf9944c8.png)

It doesn't have to be a graphical time slider, a couple of date fields with text / calendar entry will do.

egbicker:
- Current approach is going to be to add a button to the toolbar on the syslog analysis page
- Clicking this button will open a popup
- Popup will have the date entry fields and sliders
- Confirming will close the pop up then filter the data and redraw the gui as appropriate

PR #662, opened 2022-11-06
Temporal filtering of events

jw3:
Allow filtering events by a start and stop bounds.

Python API only.

Supports #465

### DIFF
--- a/crates/pyo3/src/analysis.rs
+++ b/crates/pyo3/src/analysis.rs
@@ -208,23 +208,17 @@ impl PyEventLog {
         m.into_iter().collect()
     }
 
-    fn within(&self, start: usize, stop: usize) -> PyEventLog {
-        let mut other = self.clone();
-        other.start = Some(start);
-        other.stop = Some(stop);
-        other
+    fn begin(&mut self, start: usize) {
+        self.start = Some(start);
     }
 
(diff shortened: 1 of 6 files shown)


crates/pyo3/src/analysis.rs
```diff
@@ -195,19 +221,33 @@ impl PyEventLog {
         m.into_iter().collect()
     }
 
+    fn begin(&mut self, start: Option<i64>) {
+        self.start = start;
```
jw3, line 225: @dorschs57 I was debating having these setters (`begin`, `until`) mutate state vs return a copy.  Does one way work better with the way the UI is tracking the loaded event log?

dorschs57: @jw3 I refreshed my memory on the analysis page and I think we're good with mutating it here. The use case I can see for this would be if we ever changed our implementation so you could flip back and forth between the analysis view and something else, and you wanted the temporal filter to be persisted. Mutating the globally stored state would allow for that.

dorschs57 (approved):

crates/pyo3/src/analysis.rs
jw3: 👍

PR #665, opened 2022-11-17
Filter by time range on Analysis view

### LINKED ISSUE #465: Time range selection for policy analysis

egbicker:
Adding a dialog to let user select a time range to filter the presented items in the analysis window by.

Closes #465


fapolicy_analyzer/tests/test_time_select_dialog.py
```diff
@@ -0,0 +1,57 @@
...
+    widget.get_object("ignoreStopTime").set_active(True)
+    assert widget.get_object("startMinute").get_sensitive() is True
+    assert widget.get_object("stopMinute").get_sensitive() is False
+
+
+def test_both_sensitive_toggle(widget):
+    widget.get_object("ignoreStopTime").set_active(True)
+    widget.get_object("ignoreStartTime").set_active(True)
+    assert widget.get_object("startMinute").get_sensitive() is False
+    assert widget.get_object("stopMinute").get_sensitive() is False
```
dorschs57, line 57: I'd like to see some testing of the `TimeSelectDialog.get_time` funciton. The function has a lot of edge cases where you are trapping exceptions and showing an error message that should be tested.

fapolicy_analyzer/ui/strings.py
```diff
@@ -188,3 +188,6 @@
 )
 
 APPLY_CHANGESETS_ERROR_MESSAGE = _("Error applying changes")
+
+TIME_SELECTION_START_INVALID = _("Selected Start Time is not a valid date")
+TIME_SELECTION_STOP_INVALID = _("Selected Stop Time is not a valid date")
```
dorschs57, line 193: You'll need to run the `extract_messages` script an commit the updated .pot file with the new strings.

dorschs57 (changes requested):

jw3:
I ran through this.  Some comments below on the UI, which looks like this on el8

image

1. PR needs a description.
2. The "Select" button always results in a time range around the current time, regardless of whether I previously selected something.
3. Selecting a time range does not appear to change the display.  Is this working for anyone else?
4. There is some white space at the top of the dialog
5. Title of the dialog needs set

jw3:
@egbicker 

After seeing this in action I wonder if the selection should be simpler.  What if we trim this down to a "Show events over the past X unit-of-time

So instead of picking a start _and_ stop time, just pick a start time by entering some number and selecting the unit from a drop down, eg
  - 1 hours
  - 5 minutes
  - 3 days

Some simple math to figure the seconds in whatever is selected and subtract that from time.time() and set it on the bindings, refresh the view.

We can default this to show the last hour.

The display when not in the UI can just render a string that says something like "Showing events from the last X units".  This label could event fit in the status bar, I think we talked about that option at some point, but that could also be done later if that is not very easy right now.

The button to bring up the dialog to pick the time could be moved to the toolbar.

jw3 (changes requested):
## Showing Syslog
image
- Change "Delay" to "Showing events since"
- Can the "Select" button move to the toolbar and have a clock icon?

## Showing Debug log
image
- After seeing it in action, hiding the widget rather than disabling seems appropriate. 

## Selecting a time
image
- remove the + and - buttons, just a text entry that accepts numbers

egbicker:
@jw3 Got those changes taken care of and tests updated

jw3:
> @jw3 Got those changes taken care of and tests updated

Looks better, though it doesn't seem to be working for me.  What platform are you testing the gui on?

egbicker:
Fedora but I think I might have messed something up when I was jumping back in. Mind if I give you a call?

egbicker:
Figured out the timing issue. Looks like there was a timezone issue getting in the way of correct timings

jw3:
@egbicker  looking better filtering appears to work.  I got into a weird state like this by making a selection from the subject column and then filtering in a way that eliminates that subject.

image


Also one more thing is that the previous time that was selected should be preserved in the dialog when you open it again.

egbicker:
Got the persistent data stuff working. Noticed that issue this morning, haven't gotten it working yet

egbicker:
@jw3 I think I fixed the issue we were seeing with the subjects not clearing.

jw3:
All good, just one nit.

image

Get rid of the read-only text field, just include the duration text in the label with the static text.

fapolicy_analyzer/ui/policy_rules_admin_page.py
```diff
@@ -103,6 +116,11 @@ def __init__(self, audit_file: str = None):
         )
         object_tabs.append_page(self.object_list.get_ref(), Gtk.Label(label="Object"))
 
+        self.get_object("delayDisplay").get_buffer().set_text("1 Hour")
+        self.time_delay = -1
+        self.time_unit = "2"
+        self.time_number = 1
```
dorschs57, line 122: These should probably be made "private" with the double underscore in front.

fapolicy_analyzer/ui/policy_rules_admin_page.py
```diff
@@ -500,6 +525,40 @@ def on_userTabs_switch_page(self, notebook, page, page_num):
...
+            return "s" if count > 1 else ""
+
+        time_dialog = TimeSelectDialog()
+        time_dialog.get_object("timeComboBox").set_active_id(self.time_unit)
+        buff = time_dialog.get_object("timeEntryField").get_buffer()
+        buff.set_text(str(self.time_number), len(str(self.time_number)))
+        resp = time_dialog.get_ref().run()
+        time_dialog.get_ref().hide()
+        time_unit = time_dialog.get_object("timeComboBox").get_active_id()
+        time_number = int(time_dialog.get_object("timeEntryField").get_buffer().get_text())
```
dorschs57, line 539: You could simplify a lot of this code in here if the `TimeSelectDialog` implemented getters and setters for `active_id` and `text` that hide all the implementation details.

fapolicy_analyzer/ui/policy_rules_admin_page.py
```diff
@@ -400,6 +419,12 @@ def exec_primary_data_func():
         ):
             self.__events_loading = False
             self.__log = eventsState.log
+            utc = int(datetime.datetime.utcnow().timestamp())
+            tzdelta = int(time.time()) - utc
+            if self.time_delay < 0:
+                self.__log.begin(int(time.time()) + tzdelta - 3600)
+            else:
+                self.__log.begin(int(time.time()) + tzdelta - self.time_delay)
```
dorschs57, line 427: Can you add a test to hit this branch?

fapolicy_analyzer/ui/policy_rules_admin_page.py
```diff
@@ -500,6 +525,40 @@ def on_userTabs_switch_page(self, notebook, page, page_num):
...
+
+        if time_unit == "1":
+            seconds = time_number * 60
+            display_unit = "Minute"
+        elif time_unit == "2":
+            seconds = time_number * 3600
+            display_unit = "Hour"
+        elif time_unit == "3":
+            seconds = time_number * 86400
+            display_unit = "Day"
```
dorschs57, line 549: These branches are not tested either.

dorschs57 (changes requested):

jw3 (approved):
Looks good to me.  I'll watch for @dorschs57 approval and then merge.

dorschs57 (approved):

PR #675, opened 2022-12-06
Database Sync

egbicker:
Menu option to sync trust databases after RPM install

### DIFF
--- a/fapolicy_analyzer/ui/main_window.py
+++ b/fapolicy_analyzer/ui/main_window.py
@@ -414,6 +414,9 @@ class MainWindow(UIConnectedWidget):
         if self._fapd_status != ServiceStatus.UNKNOWN:
             self._fapd_mgr.stop()
 
+    def on_syncDatabases_activate(self):
+        pass
+
     def _enable_fapd_menu_items(self, status: ServiceStatus):
         if self._fapdControlPermitted and (status != ServiceStatus.UNKNOWN):
             # Convert ServiceStatus to bool
--- a/fapolicy_analyzer/glade/main_window.glade
+++ b/fapolicy_analyzer/glade/main_window.glade
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!-- Generated with glade 3.38.2 -->
+<!-- Generated with glade 3.22.2 -->
 <!--
   Copyright Concurrent Technologies Corporation 2021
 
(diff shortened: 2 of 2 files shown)


jw3:
Binding has been merged to master, `signal_trust_reload`

egbicker:
@jw3 I'm having some issues replicating the issue. Where you installing the RPM while the application was running and which RPM were you trying to insatll?

jw3:
@egbicker I've saw this sporadically while reinstalling our app and messing with the fapolicyd install and trust database, but no solid reproducer is known.  I was thinking the tdb util could help to set up testing.. but it cant quite tweak the db the way it needs to recreate this.  Nevertheless, the function is general enough that we dont need to fix anything to need to implement it.

One addition from the additional discussion: we need to pop up a confirmation dialog after this that says we need to restart the application to see the changes.  If you have the ability to gracefully exit the application from that dialog, from say a "Restart now" button, then great.  Otherwise just make it an Ok button and they can restart at their leisure.

@dorschs57 FYI this action is eventually going to expand, but is blocked until after the trust loading stuff evolves to fix the latency issues.  When there is a mechanism  that loads trust in the background and signals the frontend of changes, then this will leverage that to reload the System trust after _this_ refresh trust function completes.

dorschs57 (changes requested):

dorschs57 (approved):

fapolicy_analyzer/ui/main_window.py
```diff
@@ -427,6 +431,25 @@ def on_profileExecMenu_activate(self, *args):
         self.__pack_main_content(page)
         self.__set_trustDbMenu_sensitive(True)
 
+    def on_syncDatabaseMenu_activate(self, *args):
+        restartDialog = self.get_object("restartDialog")
+        restartDialog.set_transient_for(self.window)
+        resp = restartDialog.run()
+        restartDialog.hide()
```
dorschs57, line 438: Can probably just call `destroy` here right?

dorschs57 (changes requested):

fapolicy_analyzer/ui/main_window.py
egbicker: There is some weird behavior that's throwing Gtk assertion errors I'm seeing with destroy:
1. Select the `Synchronize DB` menu option
2. Click `Later` or `Cancel` button
3. Select the `Synchronize DB` menu option again

dorschs57 (approved):

egbicker:
@jw3 what's missing here?

jw3:
> @jw3 what's missing here?

We are going to wait on #750 and see how this integrates with that.

1. Signal fapolicyd to reload trust db
2. Resync our trust views with disk

jw3:
I think this is going to sit until 1.1 where we can integrate this with the async trust load to eliminate the reload after notifying fapolicyd to reload.

PR #676, opened 2022-12-06
Show Rule ID on Analysis view

### LINKED ISSUE #653: Display Rule ID on Analysis view
We can jump the rule, but is it displayed any where to simply view?

dorschs57:
The rule ID associated with each object should now be shown in the 1st column.

closes #653


jw3 (approved):
:+1:

PR #677, opened 2022-12-07
Copr workflow

### LINKED ISSUE #654: Submit Copr builds through API
Copr has a REST API we could use to kick off builds, and monitor their completion, through a GitHub CI job.

https://copr.fedorainfracloud.org/api/
https://developer.fedoraproject.org/deployment/copr/copr-cli.html
https://www.mankier.com/1/copr-cli#

jw3:
Trigger Copr build from CI workflow to avoid manual initiation of builds.

Using the SRPM produced in CI to build on Copr for consistency and to avoid issues with differences between the src chroot and the build chroot.  The Copr and RPM workflows run partially parallel as the Copr workflow starts as soon as the SRPM artifact is uploaded from the RPM workflow, which is done prior to the slow RPM rebuild stage.

Supports three types of builds
- tagged releases
- latest master commit
- PR commits   

Two Copr repositories are used to provide stable and unstable streams.  The stable (fapolicy-analyzer) repo receives builds from tagged versions (eg. v1.0.0).  The unstable (fapolicy-analyzer-latest) repo receives builds from the latest commit on master.  PR commits are built using a special pr tag on the package ID which keeps the build off the public facing repo.

See the ctc-oss Copr project at https://copr.fedorainfracloud.org/coprs/ctc-oss 

Also splits the RPM artifacts up into RPM and SRPM.

Closes #654

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -5,7 +5,7 @@ OS_ID_FN = $(shell grep /etc/os-release -e "^ID=" | tr -d '"' | cut -d= -f2 | cu
 OS_ID   ?= $(OS_ID_FN)
 
 identify:
-	@echo $(shell grep /etc/os-release -e "^PRETTY_NAME=" | tr -d '"' | cut -d= -f2 | cut -d: -f1)
+	@echo "Platform $(shell grep /etc/os-release -e '^PRETTY_NAME=' | cut -d= -f2 | cut -d: -f1)"
 	@echo "Platform ID: $(OS_ID)"
 
 prepfs: prep_outdir prep_rpmbuild_dir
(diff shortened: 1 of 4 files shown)


jw3:
Might use `buildscm` if the following feature in Copr was implemented, until then we will submit srpm created in our wf.
- https://github.com/fedora-copr/copr/issues/1315 

`buildscm` might look like
```
copr-cli buildscm \
          --method make_srpm \
          --commit ${{ github.event.pull_request.head.sha }} \
          --spec scripts/srpm/fapolicy-analyzer.spec \
          --clone-url ${{ github.server_url }}/${{ github.repository }} \
          fapolicy-analyzer:pr:${{ github.event.pull_request.number }}
```

PR #691, opened 2022-12-13
Trust callbacks

### LINKED ISSUE #690: Trust checking is slow
Checking the actual status on disk with the sha256 hash can be very slow on machines with limited resources.

jw3:
Async loading of trust by implementing Python callbacks within the bindings to allow notification of progress during the trust loading process.

Alongside the callback mechanism this adds the ability to merge updated trust entries into the System without the need for a change operation.  This allows trust that has been calculated externally to be added to the System, post-init.

With these enhancements in place the default System can defer the "actual" checking, only loading the trust sources.  This greatly speeds up the load time, and removed the need for costly hashing and disk io at startup time.  Deferred loading will eventually be the default, however for the time being to avoid breaking master we will provide a binding function to load an unchecked system, `unchecked_system`.  This will be removed and the System ctor will use unchecked after the UI has integrated the callback behavior.

The "actual" checking has a new binding function `check_disk_trust` that spawns a number of threads to traverse the trust sources and hash the files on disk.  This call accepts two Python function pointers as callbacks: (1) call when a new batch of data is ready, which is passed as a list of Trust entries, (2) call when all data processing has completed.

The callback pattern appears to be reliable and supports concurrency on the backend.  There is great potential here for integration with the Redux machinery.  There is a minimal example of this in the example/async_trust.py example, which is also a good entrypoint to see the overall functionality in this PR in action.

Closes #690


jw3:
```
$ python3 examples/async_trust.py 
system created in 1.2797346115112305 seconds
   1% 1121
   2% 2242
   3% 3363
   4% 4484
   5% 5605
...
  97% 108657
  98% 109778
  99% 110899
 100% 112020
done
```

jw3:
@dorschs57 I have some refining to do yet, but this is worth taking a look at now.  The example should be a good entry point.

### tl;dr;
- System created quickly, control returns to UI
  - splash screen still useful, some systems this could take a few seconds
- Trust checking against disk is run async, callbacks used to send updates back to ui
- Passing hooks to Redux as callbacks is possible

jw3:
This is getting close.  My plan is to complete the todos and merge, @dorschs57 can integrate in separate PR.

### todo
- [ ] reuse spawned threads for multiple batches
- [ ] generalize the completion reporting
- [ ] provide total batch count
- [ ] review error handling

PR #698, opened 2022-12-15
Rust vendoring fix

### LINKED ISSUE #697: Rust tarball filtering issue
The fc rust tarball is the same size as the el tarball.

A couple spot checks confirm that there are things in it that the script output claims were unvendored.

jw3:
Fix an issue with unvendoring Rust dependencies in lock2spec.py.  This was preventing the pruning of crates available as packages from the final tarball.  The symptom was that the release artifacts were all the same size.

Closes #697


jw3:
The fc tarballs are now 428K, the el tarball is 10MB.  :+1: 

This regressed towards the end of the RPM refactor when an extra subdir needed added to the vendoring output path.

PR #702, opened 2022-12-15
Update analysis object color scheme

### LINKED ISSUE #685: Color change for Analysis Objects
In the context of the **Object column** the orange color is used for all allowed executions.

Change the color coding to represent:
- Green:  The file was trusted and the execution was successful
- Orange:  The execution was successful but the file is not listed as trusted
- Red: Denied for any reason

egbicker:
This updates the color coding in the analysis view object column to the following:

- Deny is RED
- Accept with either ST or AT is GREEN
- Accept with U is ORANGE
- Accept with an untrusted ST is ORANGE

Closes #685

### DIFF
--- a/fapolicy_analyzer/ui/object_list.py
+++ b/fapolicy_analyzer/ui/object_list.py
@@ -60,15 +60,14 @@ class ObjectList(SubjectList, Events):
         matches = set(options).intersection(valueSet)
         return seperator.join([wrap(o) if o in matches else o for o in options])
 
-    def __colors(self, access, mode):
+    def __colors(self, access, status):
         green = (Colors.LIGHT_GREEN, Colors.BLACK)
         orange = (Colors.ORANGE, Colors.BLACK)
         red = (Colors.LIGHT_RED, Colors.WHITE)
         if access.upper() != "A":
             return red
 
-        numModes = len(set(mode.upper()).intersection({"R", "W", "X"}))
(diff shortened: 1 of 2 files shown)


fapolicy_analyzer/ui/object_list.py
```diff
@@ -60,15 +60,14 @@ def wrap(x):
...
+    def __colors(self, access, status):
         green = (Colors.LIGHT_GREEN, Colors.BLACK)
         orange = (Colors.ORANGE, Colors.BLACK)
         red = (Colors.LIGHT_RED, Colors.WHITE)
         if access.upper() != "A":
             return red
 
-        numModes = len(set(mode.upper()).intersection({"R", "W", "X"}))
-        return green if numModes == 3 else orange if numModes > 0 else red
+        return green if status.lower() == "t" else orange
```
dorschs57, line 70: You also need to look at the actual trust here, not just the trust status. something like this I think:
```suggestion
        elif trust.lower() in ["at", "st"] and status.lower() == "t":
            return green
        else:
            return orange
```

dorschs57 (changes requested):

dorschs57 (approved):
This looks correct to me for the scenarios I can think to test.
* Deny is RED
* Accept with either ST or AT is GREEN
* Accept with U is ORANGE
* Accept with an untrusted ST is ORANGE

PR #713, opened 2022-12-19
Fix changesets on rollback

### LINKED ISSUE #689: Deployment after rollback loses changeset
1. Reconcile trust from analysis view
2. Deploy
3. Allow the countdown to rollback (or click no)
4. Deploy again
5. Observe that the trust was not applied

This can be done for rules

dorschs57:
@jw3 Try this one and see if it fixes the issue. I must have been thinking that a revert to the previous system brought the changesets applied to that system with it. I just reapply the changesets from the UI state to the system after its reverts.

closes #689


jw3 (approved):
This works :+1:

jw3 (changes requested):
1. Add a trust entry
2. Add and save a new rule
3. Deploy
4. Add another trust entry
5. Click Deploy
6. Observe rule changes in the deployment confirmation dialog

* The rules do get deployed to disk
* If the rule is removed after (6), it is not indicated in the next showing of the deployment confirmation dialog

dorschs57:
> 1. Add a trust entry
> 2. Add and save a new rule
> 3. Deploy
> 4. Add another trust entry
> 5. Click Deploy
> 6. Observe rule changes in the deployment confirmation dialog
> 
> * The rules do get deployed to disk
> * If the rule is removed after (6), it is not indicated in the next showing of the deployment confirmation dialog

@jw3 I was not able to reproduce this issue completely. When I did the deploys I was tailing the rules file and the trust file and both were updated. I tested both with doing a revert after step 3 and without reverting. However what I did find was another bug where the deploy confirm dialog was incorrectly reporting the rule change from the first deploy even after it was already deployed. This was just an issue with that dialog having the incorrect versions of the system and the checkpoint when it made the `rules_difference` call. This should be fixed now. If you are still seeing the issue above let me know.

PR #714, opened 2022-12-19
Unsaved Rule Indication

### LINKED ISSUE #712: Add indicator that there are unsaved rule edits
There is a place between validation and saving that is not indicated in the GUI.

1. Load rules
2. Edit rules and save
3. Edit rules again
4. Deploy rules

You have just deployed rules from (2).

An asterisk on the tab would be one simple thing.  Another might be a dialog when clicking deploy that you have unsaved changes in the editor, and an option to continue.

egbicker:
Add indications to the user that there are unsaved rule changes when going to deploy.

1. An asterisk is appended to the rule editor tab text when unsaved changes are present. 
2. A confirmation dialog is displayed when deploying with unsaved changes are present.

Closes #712


fapolicy_analyzer/ui/rules/rules_admin_page.py
```diff
@@ -121,11 +121,17 @@ def __load_rules(self):
...
+        return self.__rules_dirty()
+
     def __rules_dirty(self) -> bool:
-        return (
+        is_dirty = (
             bool(self.__modified_rules_text)
             and self.__modified_rules_text != self.__rules_text
         )
+        text = "*Rules Editor" if is_dirty else "Rules Editor"
+        self.get_object("editorView").set_text(text)
```
dorschs57, line 133: A better place for this code is in the `on_text_view_rules_changed` event handler. You just need to do your own comparison there between modified rules and original rules.

fapolicy_analyzer/ui/main_window.py
```diff
@@ -431,8 +431,11 @@ def _refresh_toolbar(self):
         self.__toolbar.refresh_buttons_sensitivity()
 
     def on_deployChanges_clicked(self, *args):
+        unsaved_rule = False
+        if isinstance(self.__page, RulesAdminPage):
+            unsaved_rule = self.__page.get_is_dirty()
```
dorschs57, line 436: I'd rather see this work done in the `DeployChangesetsOp` class instead of here. That class subscribes to the system feature so it should be able to get the information it needs to determine if the rules are dirty or not.

fapolicy_analyzer/glade/unsaved_rules_dialog.glade
```diff
@@ -0,0 +1,93 @@
...
+            <property name="position">0</property>
+          </packing>
+        </child>
+        <child>
+          <object class="GtkLabel">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="margin_top">20</property>
+            <property name="margin_bottom">50</property>
+            <property name="label" translatable="yes">Unsaved rule changes found. Would you like to continue deploying?</property>
```
dorschs57, line 78: Maybe something like "You have unsaved changes to the rules. If you deploy your changes will be lost. Would you like to continue deploying?"

dorschs57 (changes requested):

egbicker:
@jw3 I think this one is ready

jw3:
Switched the * to be at the end to avoid shifting the title

jw3 (changes requested):
1. Save a single rule `allow perm=open all : all`
2. Observe that the asterisk was removed on save
3. Edit the rule by adding some whitespace at the end `allow perm=open all : all    `
4. Save
5. Observe that the asterisk is not removed on save
6. Deploy, observe that the warning of unsaved changes is displayed

@egbicker This could be a problem that needs fixed outside this PR, can you dig into it and determine where the breakage is at

egbicker:
@jw3 I think the issue is when `dispatch(modify_rules_text)` on line 208 in `rules_admin_page.py` gets called, somewhere in the rust layer the trailing whitespace gets stripped which causes it to not show up as a different rule text string. I think maybe in the `parse` function in `/crates/pyo3/src/rules.rs` but I don't know enough about rust to say for certain.

jw3:
Thanks for the analysis @egbicker , this sounds plausible.

@dorschs57 when we deploy we replace the editor text with the updated system text.  Would that be available here?  

Think of the resulting behavior like having format-on-save turned on in a IDE.

jw3:
> I started looking into this. Hopefully I'll have something soon.

@dorschs57 , tracking in #725 
@egbicker this one will be merged

(discussed on 01-05-23 call)

fapolicy_analyzer/ui/operations/deploy_changesets_op.py
```diff
@@ -168,6 +178,18 @@ def run(self, changesets: Sequence[Changeset], system: System, checkpoint: Syste
         -------
         None
         """
+
+        rules = (
+            bool(self.__modified_rules_text)
+            and self.__modified_rules_text != self.__rules_text
+        )
+        if rules:
+            unsaved_rule_dialog = UnsavedRulesDialog()
```
dorschs57, line 187: Instead of rolling our own message dialog here we should use the built in `Gtk.MessageDialog` here. That would provide a more consistent feel to our message dialogs across the system. You can see an example of this here https://github.com/ctc-oss/fapolicy-analyzer/blob/fd5634b9df8a510afc56e6a3128173eda4815b44/fapolicy_analyzer/ui/add_file_button.py#L63

dorschs57 (changes requested):

dorschs57 (approved):
@egbicker I just pushed a few minor changes to make the dialog more consistent with how the Undeploy Changes Dialog looks when you try to close the app without deploying. Give this a look.

PR #717, opened 2022-12-21
Improved profiler env validation

### LINKED ISSUE #644: Error handling around environment parsing
1. Enter an environment that is not in `K=V` format, for example  `/home/dave/.cargo/bin`
2. Run the profiler
3. Observe silent failure with stack trace in console

```
  File "/usr/lib/python3.6/site-packages/fapolicy_analyzer/ui/profiler_page.py", line 166, in on_test_activate
    self._fapd_profiler.start_prof_session(profiling_args)
  File "/usr/lib/python3.6/site-packages/fapolicy_analyzer/ui/faprofiler.py", line 257, in start_prof_session
    self.faprofSession = FaProfSession(dictArgs, self)
  File "/usr/lib/python3.6/site-packages/fapolicy_analyzer/ui/faprofiler.py", line 63, in __init__
    dictProfTgt["envText"]
  File "/usr/lib/python3.6/site-packages/fapolicy_analyzer/ui/faprofiler.py", line 50, in _comma_delimited_kv_string_to_dict
    for k, v in dict(x.strip().split("=") for x in string_in.split(",")).items()
ValueError: dictionary update sequence element #0 has length 1; 2 is required

```

tparchambault:
Validation of Profiler's environment variable field added such that errors are presented in a manner consistent with the current profiler args validation. Variable names are also verified that they do not start with a number and can contain only [a-zA-Z0-6_] chararters.

Closes #644


tparchambault:
PR specific unit-tests still need to be generated.

tparchambault:
The env var string is expected to be a number of K=V pairs, separated by commas. A single leading double quote and a single trailing double quote will be ignored. Internal double quotes and whitespace will be untouched.

From https://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names , supported env var names are verified to support the following convention:

```
Environment variable names used by the utilities in the Shell and Utilities volume of IEEE Std 1003.1-2001 consist solely of uppercase letters, digits, and the '_' (underscore) from the characters defined in Portable Character Set and do not begin with a digit.
``` 

We can revisit supported env var names upon request. Values are not checked consequently any string is acceptable provided they don't contain commas or equal signs.

tparchambault:
New app code verifies:
1. Environment string is a comma separated list of K=V pairs
2. K is not missing, e.g. "-V" nor is empty.
3. K is comprised of only upper or lower case letters, numbers, and underscores. K can not start with a number.
4. Single leading or trailing double quotes are stripped off of V

tparchambault:
Tested rpm install over a RHEL 8.6 platform. No issues observed.

PR #732, opened 2023-01-10
Improved Python logging

### LINKED ISSUE #668: Improve error logging
Many errors encountered in the UI are presented with a toaster popup, but then there is nothing else to indicate what happened.

Need to write a log to disk.  Could be a couple of logs, one for backend and one for front, if that is significantly simpler.

tparchambault:
Several improvements to Python logging.

- Dual logging to stdout and disk
- Add log verbosity flag to cli to toggle between warn/info/debug levels
- Log points added to  debug redux dispatch events

No `v` flag specified is warn, `-v` is info, `-vv` is debug.

Closes #668


tparchambault:
The default log location is `~/.local/share/fapolicy-analyzer/fapolicy-analyzer.log.` Use the `-h` option to see the supported loglevels. 

The verbose output to stderr is now `INFO` where previously it was `DEBUG` and is not configurable.

Only a small subset of the existing `logging.debug()` statements were converted to `logging.info()`

Issue #711, opened 2022-12-19
Use trust callbacks to populate views


jw3:
Use the new async trust loading backend from #691 on the frontend.

This should allow the application to become responsive while the trust database is incrementally hashed in the background.

There is opportunity here to do some neat things
- incrementally populate the trust views
- show a trust load progress as a percentage

As noted in the #691 there is a binding call to create a System that has not had hashed checked, `unchecked_system()`.   This call can be used in place of the System constructor for now, it will eventually be removed and replace with the ctor call.

PR #738, opened 2023-01-16
Manual invocation of trust search

### LINKED ISSUE #684: System trust filter is slow
The search bar to filter the system trust list is very slow once there are any significant number of entries.

20K entries is a good number for example.

egbicker:
Addresses slowness in search due to compounding search events by changing the search bar behavior to only begin filtering when `Enter` key is pressed instead of upon any key press.

The root cause of the slowness was multiple compounding search events generated by the 150 millisecond keystroke delay.  That is now eliminated at the cost of having to press enter to initiate search.

Closes #684


fapolicy_analyzer/ui/searchable_list.py
```diff
@@ -152,6 +152,12 @@ def on_view_selection_changed(self, selection):
         data = [model[i] for i in treeiter] if model and treeiter else []
         self.selection_changed(data)
 
-    def on_search_changed(self, search):
+    def on_search_icon_press(self, *args):
+        self.get_object("search").set_text("")
         self.treeViewFilter.refilter()
         self._update_tree_count(self.__get_tree_count())
```
dorschs57, line 158: To DRY this code up a bit, make a private method called `__refilter` or something and have both handlers call it.

dorschs57 (changes requested):
Here is what I'm seeing on my system. When I type in the search box nothing happens, as expected. However when I hit enter after typing nothing happens either. I put some print statements in before the if condition on line 161 and the event key value when I hit enter is  65293. It's comparing it to the value 16777215 so nothing happens. Also when I hit enter from my number pad its a different value. Can we use the `activate` signal from the `Entry` widget instead of the `key_press` event? After a quick search online that seems to be the suggested method for how to capture an enter key from an `Entry`.

Second problem is when I type and then click the X icon to clear it, my application completely locks up to the point that I have to force close it. The text is never cleared from the search box and my application becomes completely unusable.

egbicker:
The activate is a good call. The X icon was causing me to have some unresponsive behavior prior to this branch too. If I let it sit and don't close the application, it will eventually clear the search bar.

jw3:
If there is something strange with the button and it can be removed/hidden, that would be acceptable.  A button to clear a single field is of minimal value

egbicker:
The only way I could find to remove the "clear" icon didn't seem to work before the user began inputting text, so now its getting drawn and erased on each keystroke which I don't like. Not sure what the best option here is

dorschs57:
> The only way I could find to remove the "clear" icon didn't seem to work before the user began inputting text, so now its getting drawn and erased on each keystroke which I don't like. Not sure what the best option here is

Have you tried the `icon_release` signal instead of the icon press? Its also possible things are logging up because the built in functionality of the clear button is also kicking in at the same time. You can try stopping event propagation by returning a `True` from you handler I think. You might have to handle both the press and release signals to stop propagation in both places.

egbicker:
@dorschs57 suggestion to change `SearchEntry` to `Entry` removed the clear button. Now the user must backspace and hit enter again to revert to the full trust database display. Working on the test to simulate an enter key programmatically now.

dorschs57:
@egbicker @jw3 I'm seeing some better performance from the filtering with my changes for async trust loading. Its limited testing right now, but it might be work holding off for my work on #711.

jw3:
@egbicker conflicts after #750

egbicker:
@jw3 I'm seeing an issue with the new "check_ancillary_trust" with `(left != right)`

jw3:
@egbicker Did something happen when you resolved conflicts?  I see only one line changed in this PR.

Assuming this branch is broken, I took from the previous commit and rebased on master to test.  Saw seemingly sane behavior.

List the steps to repro the error and Ill test again.

egbicker:
@jw3 I'm seeing this in my master branch, looks like one of the checks in the rust backend doesn't like what I have going. I was behind on syncing my fork so I just got that update.

jw3 (approved):
Tested this locally, looks good :+1: 

Our search is slow in general.  Need to address this, but not here.

PR #763, opened 2023-02-07
Show trust discrepancies by default

egbicker:
Show trust discrepancies by default in the STDB or a message if store is empty


jw3:
@egbicker bunch of conflicts after merging #750

egbicker:
@jw3 Is there a way to request only untrusted files from the backend?

jw3:
> @jw3 Is there a way to request only untrusted files from the backend?

There is not, right now.  Is filtering the system trust slow, or do you have some other use case?

egbicker:
I'm having issues filtering the system trust in the GUI layer while getting a performance increase. It looks like I'm going to have to chop up the `system.get("system_trust")` object before trying to load anything or else we still go through every entry.

dorschs57:
> I'm having issues filtering the system trust in the GUI layer while getting a performance increase. It looks like I'm going to have to chop up the `system.get("system_trust")` object before trying to load anything or else we still go through every entry.

Did you try doing some filtering before a trust entry is put on the queue in the `process_trust` method in `TrustFileList`. That's running on a separate thread so the performance hit from the UI standpoint shouldn't be too bad.

jw3:
@egbicker This does not seem to be functioning for me, can you take a look and confirm it is still gtg

egbicker:
@jw3 What isn't working on your end? It seems to be working on my end

egbicker:
image

jw3:
Ok, Perhaps I was just confused by these things lining up

1. Having a db with no discrepancies
2. The file count showing zero
3. Some delay on mouseover of the toggle button (it never showed to be clickable)

egbicker:
I just saw the same issue with the toggle button not becoming sensitive. I can just have it be sensitive all the time, even on the ATDB page to avoid what must be some sort of lock up since I'm only seeing it stay insensitive sometimes.

jw3:
Yeah the sensitivity of the toggle button probably needs adjusted, and needs to consider the state of the trust load.

For example, when I click toggle the first time it starts loading the trust entries into the table, and then when I click it again it doesnt seem to do anything, then if I click again it resets the trust load back to 0%.

Maybe, once you opt-in to  trust loading, the toggle button remains is disabled until it is completed?

fapolicy_analyzer/ui/database_admin_page.py
```diff
@@ -14,46 +14,64 @@
...
 import fapolicy_analyzer.ui.strings as strings
 from fapolicy_analyzer.ui.ancillary_trust_database_admin import (
     AncillaryTrustDatabaseAdmin,
 )
 from fapolicy_analyzer.ui.system_trust_database_admin import SystemTrustDatabaseAdmin
-from fapolicy_analyzer.ui.ui_page import UIPage
 from fapolicy_analyzer.ui.ui_widget import UIWidget
 
-gi.require_version("Gtk", "3.0")
-from gi.repository import Gtk  # isort: skip
```
dorschs57, line 27: This reordering of imports is odd here. Typically its best for formatter if we keep the `from gi.repository` line at the end and keep the `# isort: skip`  so formatters to try and move it out from under the `gi.require_version` check.

fapolicy_analyzer/ui/trust_file_list.py
```diff
@@ -160,7 +173,8 @@ def process_rows(queue, total):
                 return False
 
             count = self._get_tree_count()
-            if count < total:
+
+            if count < self.total:
```
dorschs57, line 177: Use `total` here not `self.total` since you are in a different thread here. Probably safe since you are only reading it, but you have the local version availble so might as well use it.

dorschs57 (changes requested):

fapolicy_analyzer/ui/trust_file_list.py
egbicker: I need to read from the other thread in order to get the final count of files with discrepancies or else we never meet the end conditions to exit and we only get the final count once everything is loaded.

jw3:
Python test failing

```
fapolicy_analyzer/tests/test_main_window.py ..
(pytest:8736): Gtk-CRITICAL **: 13:41:33.321: gtk_main_quit: assertion 'main_loops != NULL' failed
..............s..........................
fapolicy_analyzer/tests/test_notification.py ..............
fapolicy_analyzer/tests/test_object_list.py .........
fapolicy_analyzer/tests/test_policy_rules_admin_page.py ................Fatal Python error: Segmentation fault

Thread 0x00007f0ea2fc6700 (most recent call first):
  File "/home/runner/work/fapolicy-analyzer/fapolicy-analyzer/fapolicy_analyzer/ui/main_window.py", line 244 in on_next_system
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/core/observer/autodetachobserver.py", line 26 in on_next
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/core/operators/map.py", line 41 in on_next
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/core/observer/autodetachobserver.py", line 26 in on_next
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/subject/behaviorsubject.py", line 52 in _on_next_core
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/core/observer/observer.py", line 26 in on_next
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/subject/subject.py", line 55 in on_next
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/core/observer/autodetachobserver.py", line 26 in on_next
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/core/observer/autodetachobserver.py", line 26 in on_next
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/core/operators/map.py", line 41 in on_next
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/core/observer/autodetachobserver.py", line 26 in on_next
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/internal/concurrency.py", line 16 in inner
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/core/observer/autodetachobserver.py", line 26 in on_next
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/core/observer/autodetachobserver.py", line 26 in on_next
  File "/home/runner/.local/share/virtualenvs/fapolicy-analyzer-Od7Imf31/lib/python3.9/site-packages/rx/core/observer/autodetachobserver.py"Segmentation fault (core dumped)
Error: Process completed with exit code 139.
```

egbicker:
@jw3 It seems like the text display on the ATDB screen is causing some non-reproduceable issues. My local tests aren't showing this issue and sometimes the one or both of the 3.6/3.9 tests pass with no real changes.

fapolicy_analyzer/ui/trust_file_list.py
dorschs57: Ok, I see your issue there. I don't think this approach of setting the new `total` value from the `system_trust_database_admin` on load completion is the best. On my system the percent complete in the list status bar moves very slowly up to about 22% then jumps right to completed. 

I'm thinking you'll need to update the `total` in the `append_trust` method. Update it by subtracting all the entries that where not processed because they were trusted from the total every run of the `process_trust` thread (watch thread safety here). That wouldn't be perfect but it would be better. At least the total number would slowly decrease over time instead of all at once at the end. Thoughts?

fapolicy_analyzer/ui/trust_file_list.py
```diff
@@ -160,32 +175,37 @@ def process_rows(queue, total):
...
                 pct = int(count / total * 100)
```
dorschs57, line 180: May sure whatever we decide for the total you are using the same total variable here as you do in the if check on line 179.

dorschs57 (changes requested):

fapolicy_analyzer/ui/trust_file_list.py
egbicker: We still end up needing to read from the thread because we only know the initial total number of entries. I am seeing smoother behavior but its still pretty quick at the end. instead of 10% to done its, 10%, 50%, 80%, done.

jw3:
This is bumping back to 1.1.  I want more time than we have to see it in action.

jw3:
@egbicker resolve the conflicts here when you have time and I'll review this again.

egbicker:
@jw3 Merges are done, tests worked locally. Looks like something happened to the Ubuntu 20.04 repo between when you commented and now.

fapolicy_analyzer/ui/database_admin_page.py
```diff
@@ -29,31 +29,50 @@
...
-        notebook = Gtk.Notebook()
-        UIWidget.__init__(self, notebook)
-        UIPage.__init__(self)
+        self.notebook = Gtk.Notebook()
+        UIWidget.__init__(self, self.notebook)
+        actions = {
+            "toggle": [
+                UIAction(
+                    name="Toggle",
+                    tooltip="Toggle Displaying Trusted Entries",
```
dorschs57, line 38: This needs added to the `strings.py` file.

dorschs57 (changes requested):
This still doesn't seem to be working for me. On my rhel8 system, the system trust shows progress to about 15% then jumps right to done with about 14k untrusted files showing. The Toggle button never becomes active in the toolbar either. Is there something else I'm missing?

egbicker:
@dorschs57 I don't know why the toggle button isn't working, I'm seeing it respond as expected on FC34 and FC38. I did just toss up a fix for the loading jumping from low % to done.

dorschs57:
> @dorschs57 I don't know why the toggle button isn't working, I'm seeing it respond as expected on FC34 and FC38. I did just toss up a fix for the loading jumping from low % to done.

Your update made it a little better for me. However it still only gets to about 50% and then jumps to completed for me. 

I'm still seeing the issue too where the `Toggle Display Trust Entries` button does not enable after loading. This only happens when I first start start up the application. Here's my steps:
1. Start application
2. Toggle the `System Trust Database` tab
3. Wait for system trust to load complete
4. Observe the toggle button is disabled

If I then navigate to another view like Rules and back to the Trust view the button will enable properly after loading is complete.

fapolicy_analyzer/ui/database_admin_page.py
dorschs57: After your last commit this string still needs moved to the `strings.py` file.

jw3:
Ill give this a run next week and see what I see.

dorschs57:
Still seeing some issues here:
1. percentage still jumps from 55% to done when the trusted files being filtered out.
2. toggling to show all files works, but then the button is disable and I can't toggle back to just the untrusted files.
3. when I navigate to a different page my view becomes split with the trust view still showing on the top have and the page I navigated to showing in the bottom half.
image

egbicker:
Something about adding `Events` to `database_admin_page`broke the dispose condition os I fixed that but I'm no longer seeing either of your first 2 issues on fc34 or 38

jw3:
Toggle button state looks good.

Don't see the multi-view issue.

I do see the trust jump.

Peek 2023-06-15 00-27

That seems like a symptom of not reducing the total number of entries that need added to the view.

egbicker:
[Screencast from 2023-06-26 16-14-38.webm](https://github.com/ctc-oss/fapolicy-analyzer/assets/51774793/81a64d3d-ddb8-44c7-ac61-245107974081)

@jw3 Here's what my fc38 rpm loading looks like on a VM with ~8gb of RAM. I wonder if the skip you're seeing is the thread that's doing the file loading going faster than the thread responsible for redrawing the UI? Lines 229-232 in  `trust_file_list` are removing the trusted files not displayed from the total used in calculating the percentage. 

Does the loading of the full trust file database have a skip in the middle as well?

fapolicy_analyzer/ui/trust_file_list.py
```diff
@@ -200,7 +225,11 @@ def process_trust(trust, event):
             for data in trust:
                 if event.is_set():
                     return
-                self.__queue.put(self._row_data(data))
+
+                if self.show_trusted or not data.status.lower() == "t":
```
egbicker, line 229: This is where the trusted files get skipped from being displayed and the total gets decremented

PR #770, opened 2023-02-16
Profiler v2 integration

### LINKED ISSUE #767: Activity indicator for Profiler
The profiler currently blocks input while executing, but it gives no visual indication that it is active and progressing.  The combination of locked input and not visual cues can easily be confused with the application being locked up.

If we could overlay some type of message or spinner prior to blocking, that would help.

Even populating the log text area with a temporary message (that will get overwritten by the log) would be better than nothing.

jw3:
Completion of the Profiler v2 integration.

This replaces the original Python profiler implementation using a combination of Rust and Redux components. 

### Highlights
- Fully integrated with Redux
- Profiler background execution
  - Allows for a non-blocking GUI
  - Allows navigation during execution
- Terminate button kills executing process
- Supports indeterminate, ie non-terminating, processes

### Future
An existing capability that was not integrated here is the ability to profile with temporary rules, ie undeployed work in progress rules. This is implemented in the backend, but was integrated into the GUI in this PR.

Closes #767


jw3:
Coverage 95.06% => 95.81%

fapolicy_analyzer/ui/features/profiler_feature.py
```diff
@@ -0,0 +1,175 @@
...
+            return action
+
+        _handle.kill()
+
+        return terminating_profiler()
+
+    init_epic = pipe(
+        of_init_feature(PROFILING_FEATURE),
+        map(lambda _: _init_profiler()),
+    )
```
dorschs57, line 155: I don't think you need this. The epic here isn't doing any work that would change the state. I think it final state after this one would just be the default state of the `profile_reducer` correct?

fapolicy_analyzer/ui/reducers/profiler_reducer.py
```diff
@@ -13,59 +13,145 @@
...
 
 
-def handle_set_profiler_state(state: ProfilerState, action: Action) -> ProfilerState:
-    return _create_state(state, entry={**action.payload})
+def profiler_state(target, **kwargs: Optional[Any]) -> ProfilerState:
+    return target(**{**empty_profiler_state()._asdict(), **kwargs})
+
+
+def handle_profiler_init_state(state: ProfilerState, action: Action) -> ProfilerState:
+    return derive_profiler_state(ProfilerState, state)
```
dorschs57, line 101: I don't think you need this either. This action would only come from the init of the `profiler_feature` so it would always be the default state correct?

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -54,27 +56,27 @@ def __init__(self, fapd_manager):
             "start": [
```
dorschs57, line 56: We should consider adding all the analyzer related buttons to the same key in the `actions`. The top level keys here are used for grouping related buttons. If the keys are different then a divider is shown in the toolbar to signify a different group of buttons. Instead of "start", "stop" and "analyze" I'd suggest "profiler" or simular.

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -83,90 +85,152 @@ def __init__(self, fapd_manager):
...
+            ProfilerTick: self.handle_tick,
+            ProfilerDone: self.handle_done,
+        }
 
-    def on_next_system(self, system):
-        if not self.inputDict == system.get("profiler").entry:
-            self.update_field_text(system.get("profiler").entry)
+    def on_event(self, state: ProfilerState):
+        if state.__class__ in self.profiling_handlers:
+            self.profiling_handlers.get(state.__class__)(state)
```
dorschs57, line 110: 😍 this approach!

dorschs57 (changes requested):
I didn't test run this, but the code looks good other than the couple things noted.

fapolicy_analyzer/ui/features/profiler_feature.py
jw3: Yep, that's right :+1:

fapolicy_analyzer/ui/reducers/profiler_reducer.py
jw3: Correct, again :+1:

fapolicy_analyzer/ui/profiler_page.py
jw3: 10-4, updated

dorschs57 (approved):

tparchambault:
I ran into a few issues last Friday, but saw that @jw3 was still pushing changes so thought this might get addressed and wanted to test w/an updated RPM over Rhel86:

Running w/`/usr/bin/ls` copied to `/tmp/my-ls` and attempting to use the env var setting: `PATH=$PATH:.`, the profiler tool view does not update the lower pane. The image was captured after a number of minutes after invoking the profiler `Run` button. Whether the cmd was run or not is TBD however the GUI is still live and responding to mouse events. 

Note: The target's absolute path was correctly generated by the `_rel_tgt_which()` function. In the prior implementation, the target binary was always referenced via it's full path as an argument to Python's `subprocess.Popen()` constructor. 

> INFO:root:validateArgs()
DEBUG:root:validateArgs({'cmd': 'my-ls', 'arg': '-l', 'uid': 'toma', 'pwd': '/tmp', 'env': 'PATH=$PATH:.'}
DEBUG:root:Processing current working dir: /tmp
DEBUG:root:FaProfSession::validateArgs() --> pwd verified
DEBUG:root:expand_path($PATH:., /tmp)
DEBUG:root:expand_path::path = /usr/sbin:/usr/bin:/sbin:/bin:/root/bin:/tmp
DEBUG:root:exec=my-ls, Profiling PATH = /usr/sbin:/usr/bin:/sbin:/bin:/root/bin:/tmp
INFO:root:_rel_tgt_which() - return value:/tmp/my-ls
DEBUG:root:Entry text = {'cmd': 'my-ls', 'arg': '-l', 'uid': 'toma', 'pwd': '/tmp', 'env': 'PATH=$PATH:.', 'env_dict': {'PATH': '$PATH:.'}}
DEBUG:root:dispatch( START_PROFILING_REQUEST )



post_profiler_tmp_my-ls_w_path

tparchambault:
Same target except using the absolute path to the executable. Directly executing `fapolicyd --debug --permissive` sends its output to `stderr` and not `stdout` on Rhel86. Maybe this differs with a newer release of `fapolicyd`.

post_profiler_tmp_my-ls_fapd_stderr

tparchambault:
The last observation... With a specified user and unspecified working directory, the target is executed in the `root` user's home directory. The first iteration used the specified user's home directory as its `pwd` in the event the profiling target performed relative file r/w operations.

post_profiler_unspecified_work_dir_pwd_is_root

Issue #771, opened 2023-02-17
Rename debug log menu item


jw3:
The "Analyze from Audit Log" option should be renamed to "Analyze from Debug log" for correctness.  This also needs done to prevent collision with #294, which will likely add the same _"from audit log"_ menu item.

![image](https://user-images.githubusercontent.com/1545372/219663641-9ca538e3-1c44-488c-8e17-1f84f84fd7a5.png)

PR #789, opened 2023-02-23
Customizable initial application view loading from the config file

### LINKED ISSUE #773: Customize initial view
Opening to the rules view is a much better user experience as it does not immediately incur the expense of the trust load.

Thinking about it though, depending on the user's workflow, any of the views may be a good starting point.  It might be nice to allow customization of which view the app starts on.

One side effect of not starting with the trust admin view is that the initial system will not be checked.  Other views, like the analysis view, may need access to a checked trust database to provide accurate insight.  So the check still needs done somehow.

---
The following is probably a separate issue, but leaving it here for now.

What if the trust check for the initial system was a task separate from any view, that runs when the app opens.  It runs once regardless of what view the app starts on. Perhaps a single progress bar somewhere, like down in the status area could show the status of the background process.

From what I understand then the trust view would need to have progress bars, like it does now, that would show the progress of loading the trust into the table views.  These progress bars would only be indicating the status of the UI load, as the check was already completed.

The special case with that would be when the trust view is your opening view, there would be the disk check progress bar, plus the trust table progress bars.  How do they wire together?

It wouldn't even have to be the trust view is the starting view, could just be that the trust view is switched to prior to the check completing.

This relates to a few things that need reported
- Trust check continues after switching from view (UI)
- Trust check impacts focus on other views (UI)
- Trust check is repeated when switching back to trust view (UI)
- Trust check should be cancellable (Rust)

### LINKED ISSUE #804: Consistency in disabling tool menu items
When the application loads, the Trust tool is disabled.  That makes sense because trust is active.  However once switching away to a different tool the disable behavior is not applied to the new tool.

Just need to be consistent. Either remove the initial trust disabling, or make all views disabled when active.

dorschs57:
Right now this is manual process. To set your initial view update your config.toml file with a section like
```toml
[ui]
initial_view = 'trust'
```
The possible values are  "trust", "rules", "audit log", "syslog", or "profiler". With the default being "rules".

As part of this change the I removed the disabling of the trust menu option when you are on the trust screen. If the user is on the trust page and they hit the trust menu option again it just refreshes the page.

Closes #773 
Closes #804


jw3 (commented):
Will need to add a `Requires:      python3-toml` to both spec files

dorschs57:
@tparchambault Can you give the session saving and reloading a spin on this branch. I created a reusable widget to generate a file chooser dialog and switched all the places we were repeating ourselves in the code over to use it. The session management stuff used it in 3 different places.

jw3:
@dorschs57 this picked up a conflict from master

dorschs57:
@jw3 I put that minor change in to remove the + buttons.

fapolicy_analyzer/ui/features/system_feature.py
```diff
@@ -185,7 +185,6 @@ def checking_finished():
             action_fn=ancillary_trust_load_complete,
             flag_fn=checking_finished,
         )
-        print(check_ancillary_trust)
```
jw3, line 188: :+1:

fapolicy_analyzer/ui/types.py
```diff
@@ -0,0 +1,25 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+
+from enum import Enum
+
+
+class PAGE_SELECTION(Enum):
+    TRUST_DATABASE_ADMIN = "trust"
+    RULES_ADMIN = "rules"
+    ANALYZE_FROM_AUDIT = "audit log"
```
jw3, line 23: This should be "debug log".

We have had some ambiguity between debug/audit in the past.  It doesnt need fixed now other than what is user facing here.

We updated the other user facing part in #771

jw3 (changes requested):
Just the one comment on the /audit/debug/ replacement, otherwise :+1:

I tested all view configurations, including not present.  Also clicked through saving and loading sessions and a debug log.

fapolicy_analyzer/ui/types.py
jw3: @dorschs57 I did a /audit/debug/ for this

jw3:
@tparchambault any comment on this one?  I tested the pickers for the session operations, seemed to be in order.

fapolicy_analyzer/ui/types.py
```diff
@@ -20,6 +20,6 @@
 class PAGE_SELECTION(Enum):
     TRUST_DATABASE_ADMIN = "trust"
     RULES_ADMIN = "rules"
-    ANALYZE_FROM_AUDIT = "audit log"
+    ANALYZE_FROM_DEBUG = "debug log"
```
dorschs57, line 23: 👍

dorschs57 (commented):
@jw I'm good with your last commit. That was exactly what I had planned to do. If you tested it 👍

tparchambault:
Currently reviewing it...

Can the generated toml file contain a comment with the possible valid choices for the option? Not this PR but downstream...

> The possible values are "trust", "rules", "audit log", "syslog", or "profiler".

jw3:
> Can the generated toml file contain a comment with the possible valid choices for the option? Not this PR but downstream...

It could.  Though the comment in an existing config would be inaccurate if we add a new view.

User guide is probably the place for it. There is a section there now describing it.

tparchambault:
Are we adding an additional `*.toml` file or renaming the existing `fapolicy-analyzer.toml` to `config.toml`? I'm hoping the latter.

tparchambault:
Independent of our tool but something we need to be aware of, as something @jw3 also wrestled with yesterday:

> Mar 10 14:48:58 son-o-wimpy dbus-daemon[898]: [system] Successfully activated service 'org.fedoraproject.SetroubleshootPrivileged'
Mar 10 14:49:00 son-o-wimpy setroubleshoot[4401]: SELinux is preventing fapolicyd from write access on the file fapolicyd.pid. For complete SELinux messages run: sealert -l c719e1c0-b6e1-462a-b40f-ccbb8ba66612
Mar 10 14:49:00 son-o-wimpy setroubleshoot[4401]: SELinux is preventing fapolicyd from write access on the file fapolicyd.pid.
                                                  
Over RHEL8.6 via an RPM install, I was testing control of `fapolicyd` from `fapolicy-analyzer`.

tparchambault:
Only one anomaly observed and I'm assuming it's independent of this PR: After `Deploying Changesets` and allowing the `Accept these changes?` count-down dialog to expire, which initiated an automatic Trust rollback, the previous pre-deployment session state was not reloaded or at least visible in the Ancillary Trust view, (which by itself is neither good nor bad, just a design choice), however the application state remained in the `pending changes` such that attempts to quit were met with the `You have applied changes` dialog. This may be a known issue.

tparchambault (commented):
Tested over Rhel86. All good. No source comments that rise above the six of one, half dozen of another syntax and naming level. Redox and async is the future.

PR #795, opened 2023-02-25
Allow relative path user guide images

### LINKED ISSUE #756: Support relative image paths in doc gen
Allow a images directory in the wiki repo.

The overall task here should be roughly

1. Locate where the hrefs are parsed
2. Add an if statement that checks if the href starts with "http"
   - if it does, do what is done currently, that is download to output dir
   - if it does not then assume the image is relative and copy from relative path to output dir

tparchambault:
Allows for images to be stored beside the documentation in the doc repo.

Adds support for relative path image references to be used, allowing us to store images instead of relying on the GitHub content storage.  The drag and drop method to the GitHub UI will continue work with this change, though the in-repo storage is preferred.

Closes #756


tparchambault:
Performed a test where the `help:fapolicy-analyzer` prefix were omitted from the `img` `src` attributes essentially making them relative. The Gtk Webkit UI that is embedded within the `fapapolicy-analyzer` did not render the images.
The same User-Guide html file was rendered correctly by the Firefox browser on the target system.

tparchambault:
> LOCAL_HELP_GFX=1 python help update/

The magic variable to build the User-Guide.html with relative image `src` references that can be viewed with a web brower: `LOCAL_HELP_GFX` just needs to be defined within the user's environment.

tparchambault:
This PR will should copy image `src` files into the relative `media/` subdirectory. There is a possibility of basename clashes.

dorschs57:
@tparchambault take a look at the commit I just pushed. I think this will provide a better workflow. It allows for all images to be loaded and displayed from relative paths. This will work if the help doc is loaded from the local dev path or from `help` scheme path.

jw3 (approved):
Looks good :+1: 

Tested by adding an images dir to the wiki repo, adding and image and referencing  it, running the local build and verifying it displays in the html user guide.

image

is at the bottom of the guide for now.

dorschs57 (approved):
If @tparchambault is good with the changes from my last commit then 👍 Works on my machine also lol.

jw3:
@tparchambault any final words

PR #799, opened 2023-02-28
Preserve state of Rule tree views

### LINKED ISSUE #589: Preserve state of the status tree
When validating rules it is a little annoying to have to re-expand the tree each time you hit validate.

It might be nicer if the state of each branch in the tree could be preserved and reapplied post-validate.

egbicker:
Maintains the collapsed/expanded selections state of the rule list and status view when validate / save buttons are clicked.

Closes #589


jw3:
Does this do the status tree too?

image

egbicker:
It does not. It shouldn't be too much to add.

egbicker:
Status tree should preserve now

jw3 (approved):
Looks good! :+1:

jw3 (changes requested):
I was just about to merge when the restore button struck me...

Two things
1. It will for sure be confused with undo, so if it stays the icon should be changed
2. I see it does not apply to the status view, so on the rule text edit view it has no impact even though it does click.

I would say remove it. Unless a compelling case can be made to keep it.  In which case something needs done for both 1 and 2.

jw3 (approved):
looks good :+1:

jw3:
This looks good to me, will merge pending @dorschs57 review

dorschs57 (changes requested):
Could you add a couple tests against the restore functions?

dorschs57 (approved):

Issue #793, opened 2023-02-24
Update EPEL 8 Spec


jw3:
There are some lessons learned that need applied to the epel 8 spec before it is sent to the repo.

The [ripgrep rpm repo](https://src.fedoraproject.org/rpms/rust-ripgrep) is a good reference impl.

jw3:
#794 is building for 1.0 -- need to pick back into the v0.6 branch, which is what will be used in a PR to the Fedora repo

jw3:
> There are some lessons learned that need applied to the epel 8 spec before it is sent to the repo.

Minor changes and some doc.  Removing the DC build was the main task.

PR #811, opened 2023-03-09
Rust logging

### LINKED ISSUE #807: Improve Rust logging
Tie into the logging configuration created in #668.

Use the Rust log crate as the logging facade.

Use pyo3-log to bridge into the Python logging impl.

jw3:
Bridges the Rust logging into the Python logger using pyo3-log.

Closes #807


jw3:
pyo3-log will be stable on 37 tomorrow, but 38 is going to be until after the freeze, which I think is next week.  In order to get the RC out will have to disable 38.  Will do that in #793 , and merge this afterwards.

jw3:
> Will do that in https://github.com/ctc-oss/fapolicy-analyzer/pull/794 , and merge this afterwards.

Will do that in the RC release branch after this is merged.  The RC can be tested and then v1.0 is going to tag on master, where nothing will need reverted as long as this will build on the 14th.

PR #812, opened 2023-03-09
PySystem defaults to async init

jw3:
Changes the PySystem ctor to initialize as unchecked.

Provides continued support of a blocking ctor for init through addition of a new binding function `checked_system`.


jw3:
Saving for 1.1

PR #813, opened 2023-03-09
Shareable Profiler Output

### LINKED ISSUE #810: Profiler output should be copy-friendly
1. Allow text to be copied out of the output view
2. Print the arguments passed to the backend

Alternative to (1) is to add a copy button that selects all the output into the clipboard.

Use case: I want to quickly show someone what ran in the profiler without taking a screenshot.

egbicker:
Changes that make the profiler output shareable

- Allow selection and copy on the output text view
- Display the target args on the output view

Closes #810


jw3:
A couple tweaks.  The arg display was being overwritten when the profiling was complete.  Moved it to only display after the profiling was complete as a new section with the other output.

image

PR #837, opened 2023-03-24
Fix Go To Rule from debug log analysis

### LINKED ISSUE #831: Analyzer Go to Rule context menu not working from Profiler
After running the Profiler and using the Analyze Output button to load the Analyzer, the context menu in the object view to Go to Rule does not work. It does work if I load an existing log file into the Analyzer using the Open Log File button from the Analyzer page.

To Reproduce:
1. Open the Profiler tool
2. Enter `ls` as the *Executable File* argument
3. Run the Profiler
4. When it completes click the *Analyze Output* button to load the Analyzer 
5. In the Analyzer click User -> `root`, Subject -> `/usr/bin/ls` to load the Object table
6. Right click one of the objects and select *Go to Rule*
7. Observe nothing happens

egbicker:
This fixes the Go To Rule action when analyzing a debug log.

Closes #831

### DIFF
--- a/fapolicy_analyzer/ui/main_window.py
+++ b/fapolicy_analyzer/ui/main_window.py
@@ -413,7 +413,7 @@ class MainWindow(UIConnectedWidget):
         fcd.destroy()
 
     def activate_file_analyzer(self, file):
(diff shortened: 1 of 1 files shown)


dorschs57 (approved):
👍

PR #839, opened 2023-03-29
Indicate filtering in trust file count

### LINKED ISSUE #803: Trust file count should indicate displayed of total
The file count indicator should show the number of files displayed vs the total in the db.

Eg. with 5 files showing and 100 in the db it would be something like

`5 / 100`

If you want to get clever, when there are 0 files in db just show "0 files"  or if all files are displayed just show the total.  This is to avoid redundancy of  0 / 0, or 1000 / 1000.  Not a big deal though.

egbicker:
Trust file list count indicator will now display `X / total` when `0 > X < total`.

Closes #803


dorschs57 (commented):
Changes look good. It would be nice to get some test coverage on this change though.

Issue #736, opened 2023-01-11
Trust view does not consistently remove deletions


jw3:
1. Add some trust entries (or start with some)
2. Untrust some of them
3. Deploy
4. Observe that the deleted items are still present
    - this can be random; if they are not left, restart from (1)
6. With deleted leftovers present, add new trust
7. Deploy
8. Observe that the deletions are now gone.

PR #840, opened 2023-03-30
Fixed ancillary trust loading when swapping session files

### LINKED ISSUE #827: Ancillary Trust not reloaded after deployment rollback
If the Keep Changes countdown after a deployment is allowed to expire and trigger a rollback, the Ancillary Trust is not reloaded correctly in the view. If I start with an empty Ancillary Trust and add one trusted file then after rollback the list is empty and the status shows 0% loaded. If I start with files in the Ancillary Trust and add one more trusted file then after rollback the list shows the previously trusted files only. In both cases, if I select Trust from the Tool menu so the screen reloads the Ancillary Trust view reloads with the added file displayed.

To reproduce:
1. Add a trusted file to the Ancillary Trust
2. Deploy
3. Allow the Keep Changes countdown to expire and rollback
4. The Ancillary Trust view is not updated properly 
5. Select the Tool->Trust menu
6. The Ancillary Trust view refreshes with the correct files displayed

### LINKED ISSUE #828: Ancillary Trust not reloading properly after opening session file
Possibly the same issue as #827 but need to investigate to make sure. The ancillary trust view does not reload properly when switching between session files. 

To Reproduce:
1. Add some trusted files to the Ancillary Trust.
2. Save a session as session1.json
3. Add or delete some more files from the Ancillary Trust
4. Save a second session file as session1.json
5. Switch back and forth between sessions by opening one of the session files
6. The Ancillary Trust view does not fully populate with the changes from the session file and the percentage completed status gets stuck and lever finishes.

dorschs57:
Fixes some threading/callback issues when a system revert happens followed by a set of changesets being applied. This happens when a user loads a new session file with changesets in it, or the system is reverted using the post-deployment revert option. In both cases an async trust check is kicked off from the revert. When the changesets are then applied a trust check using the new System object was not initiated.

closes #828 
closes #827


dorschs57:
@egbicker can you verify these changes fix the issue you were seeing with #827?

egbicker:
@dorschs57 I can confirm this resolves #827. It has seemingly closed #736 for me as well (I haven't been able to replicate this one yet)

egbicker (approved):

jw3:
Looks like this closed #736 as well

PR #844, opened 2023-04-06
Use user HOME when profiling if user supplied and pwd empty

### LINKED ISSUE #819: Default to user home if pwd not specified
If the pwd is not specified the command should always be executed in the effective user's home.

tparchambault:
WIP: Discuss if we want to use the shell's `pwd`; That will need to be more sophisticated if the `fapolicy-analyzer` is started via the Desktop.

Fixes #819

### DIFF
--- a/fapolicy_analyzer/ui/faprofiler.py
+++ b/fapolicy_analyzer/ui/faprofiler.py
@@ -148,9 +148,20 @@ class FaProfSession:
         exec_user = dictProfTgt.get("uid", "")
         exec_pwd = dictProfTgt.get("pwd", "")
 
+        # user?
+        user_pw_record = None  # passwd file record associated with a user
+        try:
+            if exec_user:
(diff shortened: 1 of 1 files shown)


tparchambault:
Works as expected on RHEL8.6 via an rpm installation.

I suggest we eliminate the Profiler Session class if it's no longer in play or not in future plans, or rename it to something on the order of Profiler Args. We need to keep its current functionality somewhere so maybe renaming is the way to go. As an abstraction it's time may have passed. _or_ maybe it's used someplace that I'm not aware of...

tparchambault:
@jw3 or @dorschs57, the code is trivial, so whoever gets to it first. Probably one review will suffice. TIA.

jw3:
wiki https://github.com/ctc-oss/fapolicy-analyzer/wiki/User-Guide/c3982cf39ac594c6a21d2634a44bb5d7d2714643

jw3:
should have some tests

PR #849, opened 2023-04-14
Rule Deployment Difference Dialog

### LINKED ISSUE #556: Enhance the rules deployment confirmation message
Right now the on the deployment confirmation dialog, the user only gets a count of additions and deletions to the rules set. It would be nice for them to be able to view the full diff of the file. One possible option would be to add a button in the treeview row that currently displays the counts so they could popup a new dialog that displays the full diff text. This might also be part of the larger effort of converting the confirmation dialog over to a full deployment review screen like is mocked up in the wireframes.

egbicker:
Adds an option to display a full difference in the rules changeset before deployment.

Closes #556


fapolicy_analyzer/ui/confirm_deployment_dialog.py
```diff
@@ -48,6 +49,8 @@ def __init__(
 
         changes_view = self.get_object("changesTreeView")
         self.__config_changes_view(changes_view)
+        self.current_system = current_system
+        self.previous_system = previous_system
```
dorschs57, line 53:
```suggestion
        self.__current_system = current_system
        self.__previous_system = previous_system
```

fapolicy_analyzer/glade/confirm_deployment_dialog.glade
```diff
@@ -63,7 +63,22 @@
               <packing>
                 <property name="expand">True</property>
                 <property name="fill">True</property>
-                <property name="position">1</property>
+                <property name="position">2</property>
+              </packing>
+            </child>
+            <child>
+              <object class="GtkButton" id="expandButton">
+                <property name="label">Expand</property>
```
dorschs57, line 71: Since this only deals with Rules, should it be labeled something else? "Rule Details" perhaps?

fapolicy_analyzer/ui/rules_difference_dialog.py
```diff
@@ -0,0 +1,68 @@
...
+
+from fapolicy_analyzer import System, rules_difference
+from fapolicy_analyzer.ui.ui_widget import UIBuilderWidget
+from fapolicy_analyzer.ui.searchable_list import SearchableList
+
+gi.require_version("Gtk", "3.0")
+from gi.repository import Gtk  # isort: skip
+
+
+class RulesDifferenceDialog(UIBuilderWidget):
```
dorschs57, line 25: Move this file under the `rules` subdirectory with the other rules related widgets.

fapolicy_analyzer/ui/rules_difference_dialog.py
```diff
@@ -0,0 +1,68 @@
...
+               or (d.startswith("-") and d.split("-")[-1].split("\n")[0] in current_system.rules_text())
+               ):
+                continue
+
+            if d.startswith("+") and d.split("+")[-1] in current_system.rules_text():
+                new_store.append([d])
+            elif d.startswith("-") and d.split("-")[-1] in current_system.rules_text():
+                new_store.append([d])
+            elif d.startswith("-") and d.split("-")[-1] in previous_system.rules_text():
+                prev_store.append([d])
```
dorschs57, line 55: This chunk of code looks very similar to the code in `ConfirmDeploymentDialog`. Can it be abstracted out and reused? I'd suggest making it a separate method in the file, but outside of the class so it can be imported into `ConfirmDeploymentDialog` and reused.

fapolicy_analyzer/glade/rules_difference_dialog.glade
```diff
@@ -0,0 +1,108 @@
...
+            <property name="can_focus">False</property>
+            <property name="spacing">2</property>
+            <child>
+              <object class="GtkScrolledWindow" id="prevContent">
+                <property name="visible">True</property>
+                <property name="can_focus">True</property>
+                <property name="shadow_type">in</property>
+                <child>
+                  <placeholder/>
+                </child>
```
dorschs57, line 72: Add a label over this column and the `newContent` column telling the user what they are looking at. Something like "Previous Rules" and "New Rules".

dorschs57 (changes requested):
Please add some test coverage as well.

jw3:
@egbicker is this ready?

egbicker:
@jw3 I'm just working on the tests, something about getting the current and previous system states in the test setup to actually exercise these changes is proving more difficult than expected. I reached out to @dorschs57 internally but I think he might be gone for the day.

fapolicy_analyzer/ui/confirm_deployment_dialog.py
```diff
@@ -111,11 +106,11 @@ def trust_changes():
...
             ]
 
-        messages, diff = rules_changes()
+        rule_messages, rule_diff = rules_changes()
         expand_btn = self.get_object("expandButton")
-        expand_btn.set_visible(True) if [*messages] else expand_btn.set_visible(False)
+        expand_btn.set_visible(True) if [*rule_messages] else expand_btn.set_visible(False)
 
-        return ([*trust_changes()] + [*messages], diff)
+        return ([*trust_changes()] + [*rule_messages], rule_diff)
```
dorschs57, line 113: You can use a list comprehension here `[*trust_changes(), *rule_messages]`.

fapolicy_analyzer/ui/rules/rules_difference_dialog.py
```diff
@@ -22,6 +22,17 @@
...
 
+def filter_rule_diff(previous_system, current_system):
+    diffs = rules_difference(previous_system, current_system).split("\n")
+    for d in diffs:
+        if (
+           (d.startswith("+") and d.split("+")[-1] + "\n" in previous_system.rules_text())
+           or (d.startswith("-") and d.split("-")[-1].split("\n")[0] in current_system.rules_text())
+           ):
+            diffs.remove(d)
+    return diffs
```
dorschs57, line 33: I think you can add part of your if check there as well and return a tuple of `(adds, dels)`  instead single list. Something like 

```python
adds = [d for d in diffs if d.startswith("+") and f"{d.split("+")[-1]}\n" in previous_system.rules_text()]
dels = [d for d in diffs if d.startswith("-") and d.split("-")[-1].split("/n")[0] in current_system.rules_text()]
return (adds, dels)
```

Then you can call it like so `adds, dels = filter_rule_diff(previous_system, current_system)`

dorschs57 (changes requested):

fapolicy_analyzer/ui/rules/rules_difference_dialog.py
egbicker: This check is in place to handle instances where there would be an addition and a deletion detected when removing or adding a new line to the rules file where the only difference is a newline. 

The `RulesDifferenceDialog` and the `ConfirmDeploymentDialog` both use the returned list slightly differently where a list of all additions and all deletions wouldn't work because the `RulesDifferenceDialog` cares about which `system` the addition or deletion came from.

dorschs57 (approved):
👍

PR #856, opened 2023-04-19
Expand env for profiler args

tparchambault:
Expand env vars in the profiler args prior to dispatch.

- Expand $PATH.
- Conditionally expand $USER, $HOME depending on `user` field of Profiler UI.


jw3:
What's the status of this one?

jw3:
> Decided not to add $USER and $HOME expansion because would not add much to the end-user UX. Typing `/home/someuser/bin` vs `$HOME/bin` is not too much of an inconvenience.

I could see these being useful.  Say if you were profiling successive runs against different users, the HOME var expansion would prevent the need to update the pwd field.

That could be added as a separate feature, no need to put it in this bug fix.

tparchambault:
> I could see these being useful. Say if you were profiling successive runs against different users, the HOME var expansion would prevent the need to update the /home/someuser/bin param.

Fair enough. I can see that use-case improvement.

tparchambault:
Refactored the UI level profiler code. Removed faprofiler.py and associated testing. Moved its functionality into profiler_page.py. I believe we are functionally good; Consolidated expansions and path searching. Still need to revisit the unit tests, such that they are meaningful, and migrate some that were in test_faprofiler.py.

fapolicy_analyzer/ui/profiler_page.py
```diff
@@ -195,12 +229,19 @@ def get_env_text(self) -> Optional[str]:
...
-            "cmd": self.get_cmd_text(),
-            "arg": self.get_arg_text(),
-            "uid": self.get_uid_text(),
-            "pwd": self.get_pwd_text(),
-            "env": self.get_env_text(),
+            "cmd": cmd if cmd else None,
+            "arg": arg if arg else None,
+            "uid": uid if uid else None,
+            "pwd": pwd if pwd else None,
+            "env": env if env else None,
```
jw3, line 244: Why are you making these changes?

The `get_*_text()` functions already convert to optional.

jw3:
@tparchambault

1. Run profiler with `ls` executable, other fields empty
2. After profiling completed switch away from profiler and then return
3. Observe the Working Directory is set to `/root`

fapolicy_analyzer/ui/profiler_page.py
tparchambault: Valid point. This is a remnant of a change that I considered where I kept the raw values of the fields in class variables, and the expanded and processed versions of the fields also in class variables. Whitespace on the end of some fields e.g. `cmd` broke the run and/or the checks, so I may or may not need to strip() the text in field prior to assignment. I need to check this on an rpm install before going back to a single line assignment.

jw3: Ok, then it should be fine to put the `strip()` up in the `get_*_text()` functions, or even at a single point in the `_get_opt_text()` call.

tparchambault:
> > @tparchambault
> > 
> > 1. Run profiler with `ls` executable, other fields empty
> > 2. After profiling completed switch away from profiler and then return
> > 3. Observe the Working Directory is set to `/root`
> 
> This is good. Thanks for capturing this use-case. I'll look into today when I address the above.

Duplicated on FC38 Testbed w/`fapolicy-analyzer` installed via rpm.

tparchambault:
@jw3 (V1.0.2-1, observed on FC38, RHEL8.6) With the user field containing a valid user, I am seeing the working directory field populated in the Profiler Page UI form after a Profiler `Run` followed by user navigation to another tool's view and then a return back to the Profiler Page view. I'm incorrectly performing my Profiler arg variable expansions/modifications (e.g. setting defaults) on the dictionary elements that ultimately are being stored in Redux and used to refresh the Profiler form data upon returning to the Profiler view. I need to revisit the arg handling from just prior to the Profiler Page dispatch to the PyO/Rust interface that invokes the Profiling instance of `fapolicyd` 

Note: The following images were captured on a RHEL8.6 platform.
The original Profiler page view, post-Profiling run, pre-user navigation to the Trust page:

profiler_toma_pwd_null

User-navigation back to the Profiler page view, shows that setting the default working directory to that of the specified user results in the population of the Working Directory UI field:

profiler_toma_post_nav_to_trust

tparchambault:
Tested on FC38. PR addresses the use of $USER and $HOME macros in the `env` and `pwd` fields. Works as expected. Navigating away from the Profiler Page and returning (after a profiler run) maintains the users originally entered text.

Observed anomalies/idiosyncrasies  : 
1. Environment variables entered on the Profiler Page are appended to the existing pkexec environment set which are `root` user-centric, the default $USER=`root`, and $HOME=`/root` (i.e. the `root` user's HOME dir.) 
2.  $HOME and $USER are not expanded for the `arguments` field. At this point, that would be a simple modification.

PR #859, opened 2023-04-26
Analysis time button bug fix

### LINKED ISSUE #832: Analyzer Time Selection button not working
When analyzing the syslog entries, adjusting the time filter to a value less than the latest entry does not remove that entry from displaying. 
Replicate:
1. `cp /usr/bin/ls /usr/bin/my-ls`
2. Add `deny_syslog perm=open all : trust=1` to the rules file
3. start fapolicyd
4. `su` to a non-root user
5. run `my-ls`
6. Let a few minutes pass
7. Launch `fapolicy-analyzer` and `Analyze` -> `Syslog`
8. See that there is a syslog entry for `my-ls`
9. Adjust the time filter to 1 minute
10. View that the same syslog entry that is older than 1 minute is still present

egbicker:
Fixes to allow the time selection button to properly filter syslog events.

Closes #832


egbicker:
@jw3 I think there might be an issue when parsing the `when` variable in the Rust layer. I added my best guess at a print statement in my most recent commit to see the value of `e.rs.event.when` in `crates/pyo3/src/analysis.rs temporal_filter` function and got `None`. Looking at `sudo systemctl status fapolicyd.service` and `sudo journalctl` it looks like the time format is not rfc3339.

image
image

jw3:
Whats this say `grep FileFormat /etc/rsyslog.conf`

egbicker:
`module(load="biultin:omfile" Template="RSYSLOG_TraditionalFileFormat")

jw3:
Btw, this is a good find :+1: .

That format change was made way back in development of the analysis view, I had forgot about it since then, but it definitely needs to be more robust.

egbicker:
@jw3 This seems to be working after I cleared out the entries with the old format. Do you think its worth adding that grep as a check before enabling the time selection or is it better to open an issue with the parsing?

jw3:
Maybe the analyzer should warn the user if entries come back without a when entry.

You could detect that from the returned results and display a dialog that mentions the config items to check.

fapolicy_analyzer/ui/policy_rules_admin_page.py
```diff
@@ -385,6 +402,7 @@ def __populate_objects(self):
             or self.__selection_state["group"] is not None
         ):
             last_subject = self.__selection_state["subjects"][-1]
+            self.when_none = sum([True for e in self.__log.by_subject(last_subject) if e.when() is None])
```
dorschs57, line 405: Minor suggestion, but you can simplify this statement with something like this:
```suggestion
            self.when_none = any(e is None for e in self.__log.by_subject(last_subject))
```

dorschs57 (changes requested):
One minor suggested change. Otherwise the code looks good.

fapolicy_analyzer/ui/policy_rules_admin_page.py
egbicker: Thanks Shawn! Should be good to go.

dorschs57 (approved):
Code looks good 👍

Issue #775, opened 2023-02-18
Request missing crates on EPEL 9


jw3:
Did a test build on Koji for EPEL 9.  Looks like some packages that are in Fedora are not yet in EPEL 9 repo.

https://koji.fedoraproject.org/koji/taskinfo?taskID=97640494

- [x] [setuptools-rust](https://src.fedoraproject.org/rpms/python-setuptools-rust) [bz#2173215](https://bugzilla.redhat.com/show_bug.cgi?id=2173215) :x: 
- [x] ~~[rust-confy](https://src.fedoraproject.org/rpms/rust-confy) [bz#2173214](https://bugzilla.redhat.com/show_bug.cgi?id=2173214)~~
- [ ] [rust-dbus](https://src.fedoraproject.org/rpms/rust-dbus)
- [ ] [rust-libdbus-sys](https://src.fedoraproject.org/rpms/rust-libdbus-sys)
- [ ] [rust-lmdb](https://src.fedoraproject.org/rpms/rust-lmdb)
- [ ] [rust-ring](https://src.fedoraproject.org/rpms/rust-ring)
- [ ] [rust-untrusted](https://src.fedoraproject.org/rpms/rust-untrusted)

https://mirror.arizona.edu/fedora-epel/9/Everything/x86_64/Packages/r/
https://docs.fedoraproject.org/en-US/epel/epel-package-request/

jw3:
It is not possible to have setuptools-rust in epel9.

jw3:
Vendoring remains the path forward.

PR #864, opened 2023-05-03
Build for epel 9

### LINKED ISSUE #821: Add version suffix to vendor tarballs
For clarity when switching between versions.  Not sure how the lookaside cache works around different branches and different content with same names.  Unique names makes it clearer.

### LINKED ISSUE #865: Add rpm build and actions for el9
Get the spec as close to the fedora spec as possible.  Might still require a separate spec to be maintained depending on the significance of the differences.

jw3:
Adds epel 9 builds in CI and for local containers.

The recipe here is a mix between the Fedora and epel 8 builds.  There are quite a few missing epel components yet, they are being requested as part of #775.  Eventually epel 9 will closely resemble the FC build and the specs should be merged.  However for now we will maintain a separate spec like is done for epel 8.

This also adds a version suffix to the release artifacts.  The intent is to guard against human error when manually moving artifacts between GitHub and Fedora releases.

Closes #821
Closes #865


jw3:
Tested python-events from epel testing

```yaml
      - name: Enable EPEL Testing
        if: startsWith(matrix.props.dist, 'el9')
        run: |
          dnf install -y yum-utils
          yum-config-manager --enable epel-testing
```

jw3:
image

PR #869, opened 2023-05-11
Adjustable Policy Analysis Columns

egbicker:
Adjustable widths for the policy analysis screen columns to show more of the paths by default


dorschs57 (changes requested):
Minor change request above but otherwise looks good.

dorschs57 (approved):

PR #871, opened 2023-05-15
Python code reformat and add CI check

### LINKED ISSUE #798: Reformat Python code and add CI check
Running black creates changes in about 40 files, which sort of prevents it from being used.

The goal here is to force developers to run the formatter prior to committing new code.

_This doesn't need done until there is a lull in dev, ie. post v1.0, but should be done asap after that._

## The PR to fix this should look like
1. Run black to format all files
2. Add a CI check to the Python CI stage that will fail if there are any unformatted files in the commmit

dorschs57:
I turned on black formatting with the `--check` option in the CI process and added the module `flake8-black`. Linting will now report out black formatting changes too. I also formatted the python code with black and committed the changes. I get the same results from the tests before and after the changes, and a quick 👀 check looks good.

Side note you can run linting and formatting locally with `pipenv run lint` and `pipenv run format`. This part was always there.

closes #798

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -45,7 +45,7 @@ allow_prereleases = true
 [scripts]
 test = "pipenv run xvfb-run -a pytest -s --cov=fapolicy_analyzer fapolicy_analyzer/tests/"
 lint = "pipenv run flake8 fapolicy_analyzer/"
-format = "pipenv run black fapolicy_analyzer/"
+format = "pipenv run black fapolicy_analyzer/ -t py36"
 watch-test = "pipenv run xvfb-run -a pytest-watch fapolicy_analyzer/ -- --testmon -s"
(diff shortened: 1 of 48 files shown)


Pipfile
```diff
@@ -9,7 +9,7 @@ python_version = "3.6"
 [dev-packages]
 semantic-version = "2.8.5"
 setuptools-rust = "0.12.1"
-flake8 = "*"
+flake8 = "~=4.0.1"
```
dorschs57, line 12: I had to lock flake8 to the 4.x version. Trying to run anything from 5.x or 6.x on my RHEL8 dev machine results in a max depth recursion error. I'm guessing it has something to do with python 3.6.

jw3:
Nice  :+1:

PR #875, opened 2023-06-02
Add fapolicy-analyzer log and config locations to rpm

### LINKED ISSUE #768: Add config and log paths to rpm
To ensure the config and log paths are accessible after rpm install they should be added to the rpm file list.

This likely eliminates xdg directories.

- /etc/fapolicy-analyzer/config.toml
- /var/log/fapolicy-analyzer.log

tparchambault:
Update the rpm spec to use standard paths for config, log, and state files. 

Closes #768


jw3:
Resolved a conflict in `crates/pyo3/src/config.rs`

fapolicy_analyzer/tests/test_xdg_utils.py
```diff
@@ -16,105 +16,63 @@
 import os
 
 from fapolicy_analyzer.util.xdg_utils import (
-    xdg_config_dir_prefix,
-    xdg_data_dir_prefix,
-    xdg_state_dir_prefix,
+    app_config_dir_prefix,
```
jw3, line 19: There are now binding functions that provide the directory path as a string. The paths reflect whether or not XDG was enabled in the build.

- app_log_dir
- app_data_dir
- app_config_dir

tparchambault:
Thx @jw3 . Appreciate the help.

tparchambault:
Currently testing. 

Observed on an rpm install over FC38, that the session and fapolicyd archive snapshots are incorrectly going into the same directory as the log files, `/var/log/fapolicy-analyzer/`. Will test the development environment locations prior to fixing the above issue.

Integrated the new Rust api calls into the existing xdg code so that existing session tmp file and saved `fapolicyd` state (db, config, and rules files) archiving code could remain intact. There's a slightly different name convention between the Rust api directories and the xdg spec directories that I need to translate correctly.

jw3:
> There's a slightly different name convention between the Rust api directories and the xdg spec directories

Please elaborate

tparchambault:
Based on https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html, the xdg convention is that there are state, data, and config (ignoring the cache, runtime, etc) directories, and the api provides log, data, and config dirs.

So the existing code uses the xdg_state dir for both logs, and tmp files used for a restore as recommended in the xdg doc:

```
 The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:

    actions history (logs, history, recently used files, …)

    current state of the application that can be reused on a restart (view, layout, open files, undo history, …)
```

So I've just got to tweak things to separate our log file location vs our snapshot file locations.

tparchambault:
Just to capture the current behavior of this branch in the dev environment i.e. the `FEATURES` file contains the line `xdg`:
Session json files are stored in `${HOME}/.local/state/fapolicy-analyzer/`, logs are stored in `${HOME}/.local/share/fapolicy-analyzer/` and the config file is located in `${HOME}/.config/fapolicy-analyzer/`

I'll verify the rpm install does put both the fapolicyd  backup archives and the logs are placed in the same location again. I'll also look at where the session json files are stored. These get cleaned up on shutdown, and probably did not look for them during their lifetimes.

jw3:
I dont see any issues after looking through the changes.  I still need to test the RPMs.

tparchambault:
Thx @jw3 . Only tested over FC38 for both rpm installations and in the dev environment.

PR #876, opened 2023-06-05
Include build info in About dialog

### LINKED ISSUE #823: Add additional build info to packaged application
It would be useful to display additional build info on the About page.

- build environment
- build time
- git commit
- ...

Would make distinguishing between builds of same version from different source easier. Especially during testing where we have GitHub builds and Koji builds floating around.

egbicker:
Capture build information and display it in the about dialog

Closes #823


.github/workflows/rpm.yml
```diff
@@ -153,6 +153,10 @@ jobs:
           name: source0
           path: /tmp/rpmbuild/SOURCES/
 
+      - name: Generate about page information
```
jw3, line 156: This will need moved up to the source0 stage so that it gets included in the source0 tarball.

fapolicy_analyzer/ui/main_window.py
```diff
@@ -15,6 +15,8 @@
 
 import logging
 import os
+import json
+import subprocess
```
jw3, line 19: I dont think subprocess should be used for anything in this PR

fapolicy_analyzer/ui/main_window.py
```diff
@@ -358,11 +362,49 @@ def on_saveAsMenu_activate(self, menuitem, data=None):
...
+            )
+        except Exception:
+            logging.debug("Failed to query for an rpm install")
+        abspath = (
+            rpm_path
+            if not rpm_path == ""
+            else path.abspath("./data/fapolicy-analyzer-about.json")
+        )
+        if path.isfile(abspath):
+            try:
```
jw3, line 378: Either load the build-info file or show unknown.  We dont care about what is running on the current system, ie calling subprocess to get this stuff in real time is not interesting.

fapolicy_analyzer/ui/main_window.py
```diff
@@ -358,11 +362,49 @@ def on_saveAsMenu_activate(self, menuitem, data=None):
...
 
     def on_aboutMenu_activate(self, *args):
-        aboutDialog = self.get_object("aboutDialog")
-        aboutDialog.set_transient_for(self.window)
-        aboutDialog.set_version(f"v{app_version}")
-        aboutDialog.run()
-        aboutDialog.hide()
+        rpm_path = ""
+        try:
+            rpm_path = subprocess.getoutput(
```
jw3, line 367: The json file can be marked as a resource and can be looked up like a glade file, we do not have to look it up with a call to rpm.

The build-info file can be wired into the setup.py like the glade files are, and because there is a placeholder build-info file, the setup.py build will work from both a dev-env and rpm install.

data/fapolicy-analyzer-about.json
jw3, line 1: Rename this file `build-info.json`

scripts/about.py
jw3, line 1: Rename this file `build-info.py`

scripts/about.py
```diff
@@ -0,0 +1,43 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import json
+import subprocess
+from os import path
+
+
+abspath = path.abspath("./data/fapolicy-analyzer-about.json")
+os_info = subprocess.getoutput(["uname -nr"])
```
jw3, line 22: It would be easier to test this script if you use argparse and set these from args rather than calling subprocess from here.

jw3 (changes requested):

.github/workflows/rpm.yml
```diff
@@ -51,6 +51,10 @@ jobs:
           ref: ${{ github.ref }}
           fetch-depth: 0
 
+      - name: Generate build info
+        run: |
+          /bin/python3 scripts/build-info.py data/build-info.json
```
jw3, line 56: Could add this command to the root level Makefile under target named `build-info`.

Then in CI and our dev envs we can say `make build-info`

jw3:
Looking pretty good.  What remains?

data/build-info.json
```diff
@@ -0,0 +1,5 @@
+{
+    "os_info" : "OS_UNKNOWN",
```
jw3, line 2: I think these could default to "unknown" with intent of being human readable.

_(see related comment on the build-info.py script)_

scripts/build-info.py
```diff
@@ -0,0 +1,53 @@
...
+git_info = subprocess.getoutput(["git log -n 1"])
+time_info = subprocess.getoutput("date")
+
+
+if path.isfile(args.path):
+    try:
+        f = open(args.path, "r")
+        data = json.load(f)
+        f.close()
+        if "OS_UNKNOWN" in data["os_info"]:
```
jw3, line 36:
```python
if "OS_UNKNOWN" in data["os_info"]:
```

Since this is the script that will be run in dev env or CI, I dont think you need to check and see if a value has already been set.  Should be safe to blindly overwrite with the results of the subprocess calls.

scripts/build-info.py
```diff
@@ -0,0 +1,53 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import json
+import subprocess
```
jw3, line 17: I said earlier that subprocess shouldnt be in the PR, that was incorrect. Your usage here is fine.  I should have said it shouldnt be used at runtime, only at build time.

:+1:

fapolicy_analyzer/ui/main_window.py
jw3: > I dont think subprocess should be used for anything in this PR

I was too broad with this, it shouldn't be used for anything at runtime.  Build time is good, which is what is being done currently.  :+1:

egbicker:
> 

I'm still not seeing the script outputs in my RPM build in fc38 even when putting the command up in source0

jw3:
> I'm still not seeing the script outputs in my RPM build in fc38 even when putting the command up in source0

Ok, Ill work up an example.

jw3:
See this build and branch  https://github.com/ctc-oss/fapolicy-analyzer/actions/runs/5358449384/jobs/9720893621#step:8:94

Treating your json as a resource is the key. I mentioned it in a comment above, should have elaborated a little more.

fapolicy_analyzer/ui/main_window.py
```diff
@@ -358,11 +362,35 @@ def on_saveAsMenu_activate(self, menuitem, data=None):
...
+        }
+
+        if rpm_path:
+            try:
+                f = open(rpm_path, "r")
+                data = json.load(f)
+                f.close()
+
+            except Exception:
+                logging.debug("About JSON could not be opened")
```
jw3, line 380: > About JSON ...

"The build info json" ...

.github/workflows/rpm.yml
```diff
@@ -51,6 +51,10 @@ jobs:
           ref: ${{ github.ref }}
           fetch-depth: 0
 
+      - name: Generate build info
+        run: |
+          /bin/python3 scripts/build-info.py fapolicy-analyzer/resources/build-info.json
```
jw3, line 56: `make build-info`

fapolicy_analyzer/ui/main_window.py
```diff
@@ -358,11 +362,35 @@ def on_saveAsMenu_activate(self, menuitem, data=None):
...
+            "os_info": "Unknown",
+            "git_info": "Unknown",
+            "time_info": "Unknown",
+        }
+
+        if rpm_path:
+            try:
+                f = open(rpm_path, "r")
+                data = json.load(f)
+                f.close()
```
jw3, line 377: Dont f.close, use a managed context with `with`

scripts/build-info.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+if path.isfile(args.path):
+    try:
+        data = {
+            "os_info" : os_info,
+            "git_info" : git_info,
+            "time_info" : time_info,
+        }
+        f = open(args.path, "w")
```
jw3, line 38: Same as above, use `with` for a managed context

fapolicy_analyzer/resources/build-info.json
jw3, line 1: Accidental commit...

1. Revert this
2. Commit
3. Add the relative path to the build-info file to the root .gitignore
4. Commit

fapolicy_analyzer/resources/build-info.json
```diff
@@ -1,5 +1,5 @@
 {
-    "os_info" : "OS_UNKNOWN",
```
jw3: I think these values should be set to "Unknown" as default in here too.

jw3:
It hit me a little late that the git archive call used to vendor the source is not going to package the patched file.

Moving the `make build-info` call into the spec will be a better way to handle this anyways.

jw3:
Took some time to think more on this and the build info can't be taken from just one place.  It's an aggregation of steps along the way.

At a minimum it is two places

1. Commit info while in GitHub CI
2. Build platform info during rpm build (within spec)

So the build-info.json is going to have keys added to it along the way, the build-info.py script will need to accept args that it inserts to the existing file.

I pushed a few commits that move things in this direction.  See how that looks.

egbicker:
@jw3 This looks good, I fixed the test/dev usage for it but its working on my fc38 rpm build

jw3:
I am going to run rc1 through fedora and see what the build info looks like over there before this merges.

jw3 (approved):

jw3:
Restored the use of the original about dialog

image

Added the build info in the comment section.

PR #883, opened 2023-06-29
Green text hard to read when selected

egbicker:


jw3:
The item selection seems to have slowed significantly with this implementation.  For single clicks its not a big deal, if you scroll down through the list using arrow down key you can see the jumping and skipping.  I guess this is caused by the additional event logic.

Wonder if there is any way to style the selection coloring with css...

PR #887, opened 2023-07-10
855 rule deploy override

egbicker:


egbicker:
image
Here's the current version of the dialog that pops up when saving an invalid set of rules

Issue #891, opened 2023-07-15
Python CI failing on pipenv keep-outdated flag


jw3:
```
pipenv.patched.pip._vendor.packaging.requirements.InvalidRequirement: Parse error at "'--keep-o'": Expected W:(0-9A-Za-z)

...

pipenv.vendor.requirementslib.exceptions.RequirementError: Failed parsing requirement from '--keep-outdated'
```

`Successfully ... pipenv-2023.7.11`

https://pipenv.pypa.io/en/stable/changelog/#pipenv-2023-7-9-2023-07-09

PR #892, opened 2023-07-15
Pin pipenv to 2023.7.4

jw3:
This avoids CI breakage due to removal of the deprecated `keep-outdated` cli flag.

This will be revisited, the issue will remain open.

See #891

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


.github/workflows/python.yml
```diff
@@ -109,7 +109,7 @@ jobs:
         id: install_deps
         run: |
           python -m pip install --upgrade pip
-          pip install pipenv
+          pip install pipenv==2023.7.4
```
jw3, line 112: This site does not use the keep-outdated flag, updating anyways for consistency.

Issue #886, opened 2023-07-06
Parsing backend for fapolicyd configuration


jw3:
Parsing support for fapolicyd configuration files that validates the syntax of the configuration file and provides errors through bindings for use in UI.

PR #897, opened 2023-08-10
Static load config

jw3:
Temporary unvalidated config loading mechanism.

#886


jw3:
See the included example for the basic usage.

PR #901, opened 2023-08-11
XDG feature

### LINKED ISSUE #900: Add feature to enable xdg in dev env
Use a cargo feature to switch between XDG and standard locations.  This will allow opt-in to XDG for dev environments.

jw3:
Move XDG functionality behind a cargo feature that disables it by default.

The intent here is to remove XDG support from deployment, while keeping it as an option for development environments.

Without XDG enabled the following are used:
- `config_dir`: /etc/fapolicy-analyzer
- `data_dir`: /var/lib/fapolicy-analyzer
- `log_dir`: /var/log/fapolicy-analyzer

To use XDG in a development environment ensure the `FEATURES` file is present in the root of the workspace and add a line for `xdg` to it.  This file is gitignored, which makes it the most convenient way of configuring XDG per workspace.

Closes #900


jw3:
Behavior exhibited after rpm install is that the `/etc/fapolicy-analyzer` dir will be created and used for configuration, while the logs and data will remain in XDG land.  This is because the Python side is not updated in this PR, and the data and logs are invoked from Python.

The remainder will be resolved in #875 or another supporting PR.

PR #909, opened 2023-09-11
Add the daemon config UI

### LINKED ISSUE #830: Add GUI support for managing fapolicyd.config
Text editor similar to rules for editing and validating fapolicyd configuration.

egbicker:
Adds the user interface components for fapolicyd configuration editing and deployment.

Closes #830

### DIFF
--- a/fapolicy_analyzer/ui/changeset_wrapper.py
+++ b/fapolicy_analyzer/ui/changeset_wrapper.py
@@ -57,6 +57,7 @@ class Changeset(ABC, Generic[T]):
 
             except TypeError:
                 print("Invalid changeset type to deserialize")
+        raise TypeError("Invalid changeset type to deserialize")
 
 
 class ConfigChangeset(Changeset[str]):
--- a/fapolicy_analyzer/ui/changeset_wrapper.py
+++ b/fapolicy_analyzer/ui/changeset_wrapper.py
@@ -46,7 +46,7 @@ class Changeset(ABC, Generic[T]):
                 rcs = RuleChangeset()
                 rcs.parse(data)
                 return rcs
-
+            except TypeError:
                 try:
                     ccs = ConfigChangeset()
                     ccs.parse(data)
@@ -54,10 +54,7 @@ class Changeset(ABC, Generic[T]):
 
                 except TypeError:
                     print("Invalid changeset type to deserialize")
-
-            except TypeError:
-                print("Invalid changeset type to deserialize")
-        raise TypeError("Invalid changeset type to deserialize")
+                    raise
 
 
 class ConfigChangeset(Changeset[str]):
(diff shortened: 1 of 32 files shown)


egbicker:
@jw3 I'm trying to implement the `ConfigChangeset` but even after rebuilding the rust layer I'm still getting `AttributeError: module 'fapolicy_analyzer' has no attribute 'ConfigChangeset'`

jw3:
> @jw3 I'm trying to implement the `ConfigChangeset` but even after rebuilding the rust layer I'm still getting `AttributeError: module 'fapolicy_analyzer' has no attribute 'ConfigChangeset'`

I missed a registration call in #908.  #910 adds it and adds an example config changest usage.

jw3:
Deployment nearly working, needs #911

jw3:
@egbicker do you have tests coming for the new stuff?

egbicker:
@jw3 Just got them pushed

fapolicy_analyzer/ui/changeset_wrapper.py
```diff
@@ -42,13 +42,38 @@ def deserialize(cls, data: Union[Dict[str, str], str]) -> "Changeset":
...
         elif isinstance(data, str):
-            rcs = RuleChangeset()
-            rcs.parse(data)
-            return rcs
+            try:
+                rcs = RuleChangeset()
+                rcs.parse(data)
+                return rcs
 
+                try:
```
jw3, line 50: Whats going on here?

jw3 (commented):
This is looking pretty good. A few things

- [ ] Noted the nesting exception handling above for clarification
- [ ] I still saw the unsaved rule notification, but that looks like a problem with the rules view state management
- [ ] Can a summary of config changes be added to this view?  What bindings do we need added for this?
image

jw3:
Added #912 describing some related concerns.  Doesnt have to be fixed for this PR, but could be if simple enough.

Another thought with this is that the unsaved rule edit check would apply to config as well.

fapolicy_analyzer/ui/changeset_wrapper.py
egbicker: It's trying to load the changeset data and now we have 2 `Changeset` that return as string and the `RuleChangeset` has additional exceptions in the rust layer (I think its something about having 0 rules)

egbicker:
@jw3 I believe we'd need a binding like `rules_difference`

fapolicy_analyzer/ui/changeset_wrapper.py
egbicker: It was but not like that. Good catch.

jw3:
#913 is on the way

PR #914, opened 2023-09-21
Daemon state preserved on session load

### LINKED ISSUE #833: Daemon state should not change when loading a user session json file
On RHEL86, V1.0.0, installed from EPEL-testing, with the `fapolicyd` stopped, after successfully loading (but not deploying) a saved session file, `fapolicyd` was started without any explicit user action, i.e. automagically started the daemon.

Use of our tool should not introduce side-effects.

tparchambault:
Modified code so that json session file operations are only applied to the current system object's state. No rollbacks or clearing of either the current system object maintained in the UI layer nor fapolicyd's state occur.

Closes #833

### DIFF
--- a/crates/app/src/app.rs
+++ b/crates/app/src/app.rs
@@ -11,6 +11,7 @@ use serde::Serialize;
 use std::path::PathBuf;
 
 use fapolicy_analyzer::users::{read_groups, read_users, Group, User};
+use fapolicy_daemon::conf::ops::Changeset as ConfigChanges;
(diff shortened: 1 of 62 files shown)


tparchambault:
I would like to verify that application restores post-crash are correct prior to merging.

tparchambault:
Tested via rpm install over FC38. Simulated crash via `# kill -p <pid>` and observed expected behavior. Existing session file was detected on start-up, and the user was presented with a dialog providing them with the option to restore their prior session.

jw3 (approved):
:+1:

PR #920, opened 2023-10-04
Reversed order of probable causes in the initialization dialog

### LINKED ISSUE #919: Change the order of possible issues in start-up initialization failure dlg
We provide a list of possible issues to investigate if the fapolicy-analyzer doesn't complete initialization correctly. Since we don't explicitly specify our dependency on `fapolicyd` in the rpm spec file, the installation of `fapolicyd` should be the first thing the user verifies, and our troubleshooting suggestions should consequently list that first.

tparchambault:
Changed ordering of the first two hints in the initialization failure dialog.

We are now starting with the daemon installation as it is logically the first thing to check on.

Closes #919


jw3:
Looks good to me

image

PR #921, opened 2023-10-05
Notify user of unsaved edits

### LINKED ISSUE #912: Text editor state issues
The text editor views (rules, config) allow the user to navigate away from the editor while unsaved changes are present.  When the user navigates back the unsaved changes are gone.

The system does however appear to track some notion of the missing state, this is evident from the rules editor

1. Edit rule text
2. Navigate away without saving
3. Attempt to deploy from another view
4. Observe the unsaved rules dialog

image

egbicker:
Notify user of unsaved text when navigating away from editor.

The user can decide whether to remain on the page, or discard unsaved changes to navigate away.

Closes #912


jw3:
Config dialog has rule text

image

PR #924, opened 2023-10-13
Fix required gtk version in config editor glade files

### LINKED ISSUE #923: Fix Gtk version in fapolicyd config file editor UI elements
We are using an older Gtk minimum version to support RHEL 8.

tparchambault:
This fix pins the gtk version to 3.22 in the `fapolicyd` config editor glade files.

The config file editor was broken on RHEL8 at the 3.24 version of Gtk.

Closes #923


tparchambault:
Verified that the `fapolicyd` config file editor works over FC38, FC39, RHEL8.6, and RHEL9.2 (rpm installs).

PR #928, opened 2023-10-19
Erroneous Unsaved Dialog 

### LINKED ISSUE #927: Erroneous unsaved changes prompts
Looks like #921 may have included an issue with prompting for unsaved changes even when there are no unsaved changes.

1. Edit rules on config
2. Save
3. Navigate away
4. Observe prompt

egbicker:
Fixes notifications about unsaved changes that do not exist.

Closes #927


jw3:
@egbicker the rules view is fixed, but the config view still seems to have the same problem.

jw3:
@egbicker the notification occurs only when there is invalid config in the editor.  Follow up PR will address the lack of notification of syntax error.

Issue #929, opened 2023-10-20
Fc not building due to missing crate


jw3:
```
error: failed to select a version for the requirement `directories = "^4.0"`
candidate versions found which didn't match: 5.0.1, 3.0.2
```

https://github.com/ctc-oss/fapolicy-analyzer/actions/runs/6574210862/job/17859188326#step:8:1946

jw3:
Seems to exist https://mirrors.kernel.org/fedora/releases/38/Everything/source/tree/Packages/r/rust-directories-4.0.1-3.fc38.src.rpm

jw3:
Fc 37 has same issue

jw3:
https://lists.fedoraproject.org/archives/list/devel-announce@lists.fedoraproject.org/thread/C6U5YUK5RPPWIWABKDVJYBSFOJ7RSOSN/

Delayed to 10-31-2023

jw3:
0.5.0 will soon be on all platforms, going to bump crate to 0.5

https://src.fedoraproject.org/rpms/rust-directories

jw3:
Closed in #945

jw3:
Fedora 37 will go EOL on 2023-12-05

PR #946, opened 2023-11-07
Update fc builds

jw3:
Drops fc37, adds fc40, restores rc39.

#929


jw3:
Fedora 37 will go EOL on 2023-12-05

PR #960, opened 2023-12-11
Add semver check using defaults

jw3:


jw3:
Testing out an enum change

image

PR #990, opened 2024-01-10
Reload rules while profiling

### LINKED ISSUE #985: Hash rules db
Provide a sha256 of the compiled rules db

### LINKED ISSUE #989: Reload rules in live profiler
When the profiler is running the current ruleset can be deployed and dynamically reloaded using the piped reload from #877.

This should not perform a full system deployment.

The cleanup from the rule write is free, provided by the normal flow of the profiler shutdown.

jw3:
Allow rules to be loaded dynamically into a profiling session.

This allows for a better test / update cycle while profiling as the profiler does not have to be shutdown to update rules.

This also adds a new concept, the "Rule Identity", which provides a sha256 hash of the rule database to allow change monitoring across both profiler and systemd daemon execution. This hash is based on the compiled rules to provide a precise content based hash.

Closes #985 
Closes #989


fapolicy_analyzer/ui/main_window.py
```diff
@@ -197,7 +197,7 @@ def __pack_main_content(self, page: UIPage):
                 page.dispose()
                 return
 
-        if self.__page:
+        if self.__page is not None:
```
jw3, line 200: Cause of split view when navigating away from rule view after Events was mixed in.

The type after Event mixin must not have had a __bool__ ability.  It also didnt have a string rep, perhaps that was one in the same with the bool rep.

Checking for `not None` here is appropriate regardless, since there is not a valid bool rep of a page.

PR #1009, opened 2024-03-05
Add demo pipeline source and wrapper

tparchambault:
This is the source to a six process pipeline that passes data via files in shared directories. This example is intended for demonstrating fapolicy-analyzer Profiler's ability to handle more complex applications, and present the results of the associated Profiler session.


tparchambault:
I will add License and CTC headers tomorrow AM.

PR #1014, opened 2024-06-03
Trust Filter UI

### LINKED ISSUE #758: Support rpmdb filtering config
fapolicyd has now externalized the rpmdb filtering config
- https://github.com/linux-application-whitelisting/fapolicyd/pull/222
- https://github.com/linux-application-whitelisting/fapolicyd/issues/151

This feature gives us the support that was previously prototyped and discussed at 
- https://github.com/ctc-oss/fapolicy-analyzer/discussions/275

This new feature will be supported in a couple ways
1. A rpm filter config tool, which will configure explicit inclusions/exclusions.
2. Various entrypoints for "Untrust" behavior on system trust entries
    - System trust view
    - Analysis view on entries that are only ST
 
There should be no need to support a legacy version in the analyzer.  The rpmdb filtering is currently only used in the tdb utility, which has no stability guarantees at this point.

jw3:
Add UI components to edit fapolicyd trust filter.

Closes #758

### DIFF
--- a/crates/app/src/app.rs
+++ b/crates/app/src/app.rs
@@ -6,10 +6,10 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  */
 
+use std::path::PathBuf;
+
 use serde::Deserialize;
 use serde::Serialize;
-use std::iter::Filter;
-use std::path::PathBuf;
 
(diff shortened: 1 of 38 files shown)


jw3:
Rebased on #1016 to fix build issues

PR #1037, opened 2024-10-07
Fix trust init error

jw3:
Investigating an issue with trust init reported in #1036...


jw3:
Moving branch to ctc-oss so Copr build will run on PR..

PR #1044, opened 2024-11-10
Backwards compat for webkit version

### LINKED ISSUE #1031: Fix webkit version in require_version
In fapolicy_analyzer/ui/help_browser.py the require remains at 4.0, which only impacts rawhide, but is significant impact.

Will patch for now in the rpm repo, fix here in 1.4.1.

jw3:
This is needed due to difference in webkit between Fedora and EPEL.

There is a patch in the RPM repo that was added when this was discovered, that will disappear when the next package is merged.

Closes #1031


jw3:
Should be checked on Rawhide and EPEL 9

